Issue #324, opened 2026-07-03
Race: workflow transition may not observe re-anchoring completion


devtank42:
## Severity: MEDIUM

The finalize guard reads the pending-placement count outside a lock that also covers the transition, so a concurrent re-anchoring completion/regression can be missed.

**Evidence:** `qnop-core/src/main/java/io/qnop/service/review/ReviewWorkflowService.java:144-150`

**Fix:** pessimistic lock (`@Lock(PESSIMISTIC_WRITE)`) on the document, or combine guard+transition in one query.

_Surfaced by the multi-role code review of `main`@72f1fb7 (2026-07-04)._

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

PR #505, opened 2026-07-17
feat(teams): let team leads self-manage their own teams (#470)

### LINKED ISSUE #470: Team management: let team leads self-manage their own teams
## Summary

Let **team leads** manage their own teams without being a global `ADMIN`: add/remove members and promote/demote other leads — for the team(s) they lead only. Today the `TeamRole.LEAD` value exists and is documented as *"Manages the team and its membership; the team's point of contact"*, but **it is not wired to any authorization** — every team operation is `ADMIN`-only. This issue makes `LEAD` a real capability, backend and UI.

## Current state

- **Model is ready.** `Team` + `TeamMembership` (`team_role` ∈ `{LEAD, MEMBER}`) exist; `team_membership` has `ux_team_membership_team_user` and an `ix_team_membership_user` index (so "teams I belong to / lead" is index-supported). `TeamRole.LEAD` is defined but **never read for any decision** anywhere in `src/main`.
- **All team management is admin-only.** `TeamService` is a plain CRUD service with **zero authorization logic**; authz is enforced purely by the URL rule `/api/v1/admin/** → hasRole('ADMIN')` (`SecurityConfiguration`). Every endpoint lives under `/admin/teams` (`TeamController implements AdminTeamsApi`): list/create/get/update/delete + `addTeamMember` / `setTeamMemberRole` / `removeTeamMember`.
- **No lead surface exists.** The only non-admin team endpoints are read-only principal search (`GET /principals`, `GET /principals/teams/{teamId}/members`). There is no "my teams" endpoint and no non-admin write path.
- **The SPA has no signal that a user is a LEAD.** `GET /users/me` returns only the global `UserRole`; `authStore` stores only `me.role`. Frontend team management is entirely in the admin area (`pages/admin/TeamsPage`, `TeamDetailPage`), and every `useTeams.ts` hook calls the generated `adminTeamsApi`.
- `TeamMembershipRepository` has no `findByUserId` / lead-counting query yet.

## Goal

A user who is `LEAD` of team X can, for team X only: view it, add members, remove members, and change a member's team role (promote to `LEAD` / demote to `MEMBER`) — through a non-admin "My Teams" surface. A `MEMBER` (or a non-lead) cannot. `ADMIN` retains full cross-team control (create/rename/delete/list) as today.

## Scope

**Backend**
- Add `TeamMembershipRepository.findByUserId(...)` (+ a lead check such as `existsByTeamIdAndUserIdAndTeamRole(teamId, userId, LEAD)` and a lead-count query for guardrails).
- Add a **lead-scoped authorization** layer (new — none exists today): a `TeamService.isLead(teamId, userId)` guard (or a method-security `@PreAuthorize` permission evaluator; `@EnableMethodSecurity` is already on) applied to the new lead endpoints. `ADMIN` short-circuits the guard.
- Add **non-admin endpoints** (the `/admin/**` rule blocks leads, so these are new paths under `.anyRequest().authenticated()` with explicit in-handler authz), OpenAPI-first:
  - `GET /teams/mine` — teams the caller belongs to, flagged with the caller's team role.
  - `GET /teams/{teamId}` (lead-or-admin), `POST /teams/{teamId}/members`, `PATCH /teams/{teamId}/members/{userId}`, `DELETE /teams/{teamId}/members/{userId}` — each gated to a LEAD of that team (or ADMIN). Reuse the existing `TeamService` member operations behind the guard.
- **Guardrails:** a team must never reach **zero leads** (block demoting/removing the last lead — see the related last-lead concern in the user-removal issue), and prevent a lead from locking themselves out inadvertently (mirror the admin self-lockout guard). Renaming/creating/deleting teams stays `ADMIN`-only.

**Frontend**
- Expose lead status to the SPA — either extend `/users/me` with the caller's led-team ids or drive a "My Teams" surface off `GET /teams/mine`; add the corresponding `authStore`/query signal (the existing `RoleRoute`/`navConfig` gate on the **global** role only, so lead-gating needs this new per-team signal).
- A **"My Teams"** page + route, visible in the nav only when the caller leads at least one team, with member management (add/remove, role change). Reuse `AddMemberDialog`, `TeamRoleBadge`, the member table, and the `PrincipalsController` search for picking users — but pointed at **new non-admin hooks** (the current `useTeams` hooks are hard-wired to `adminTeamsApi` and can't be reused as-is for a non-admin lead).

**Tests**
- Backend: LEAD of team X can add/remove/promote within X (200) but not in team Y (403); MEMBER → 403; ADMIN → 200 everywhere; last-lead and self-lockout guards enforced.
- Frontend: nav/route visible only to leads; member-management actions call the non-admin client and reflect results.

## Out of scope
- Team creation/renaming/deletion by leads (stays admin).
- Nested/hierarchical teams; team-scoped document permissions.

## Acceptance criteria
- [ ] A LEAD can manage **only their own** team's membership and roles via non-admin endpoints; MEMBER → 403; ADMIN unaffected.
- [ ] `GET /teams/mine` (or an equivalent) tells the SPA which teams the caller leads.
- [ ] A "My Teams" page lets a lead add/remove members and promote/demote leads; it is hidden from non-leads (nav + route guard).
- [ ] Last-lead and self-lockout guardrails prevent a team from losing all leads.
- [ ] Backend + frontend gates and guardrails covered by tests; backend gate and `pnpm` frontend gate green.

## Notes
- Overlaps with the user removal/deactivation issue on **last-lead protection** — coordinate so the guard lives in one place.
- Reuse: `TeamService` CRUD, `TeamMembership`/`TeamRole`, `PrincipalsController` search, `RoleRoute` pattern, existing team React components.

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

devtank42:
## Summary

Closes #470. Lets a team **LEAD** self-manage the team(s) they lead — add/remove members and promote/demote leads — through a new non-admin **"My Teams"** surface, without being a global `ADMIN`. An `ADMIN` still passes for any team; a `MEMBER` or non-lead gets `403`. Team creation, renaming and deletion stay `ADMIN`-only under `/admin/teams`.

Until now `TeamRole.LEAD` existed but was wired to no authorization — every team operation was `ADMIN`-only. This makes `LEAD` a real capability, backend and UI.

## Backend (OpenAPI-first)

- New `Teams` tag with non-admin paths, each gated **in the handler** (the `/admin/**` URL rule doesn't cover them):
  - `GET /teams/mine` — the caller's teams, flagged with their team role.
  - `GET /teams/{teamId}` · `POST /teams/{teamId}/members` · `PATCH|DELETE /teams/{teamId}/members/{userId}` — LEAD-of-that-team (or ADMIN) only.
- `TeamService` gains a lead-scoped guard (ADMIN short-circuits) plus guardrails: a team can **never lose its last lead** through this surface (`409 LAST_LEAD`), which also blocks a sole lead's self-lockout. New `TeamMembershipRepository` derived queries back the lead check, the last-lead count and the "leads any team" flag; `/teams/mine` reuses the existing `findTeamsOfUser`.
- `GET /users/me` now returns `teamLead` so the SPA gates the surface without a separate round-trip.
- `TeamAccessForbiddenException` (framework-light) → `403`, keeping `qnop-core` free of `spring-security-web` (same pattern as `AnnotationActionForbiddenException`; no `@PreAuthorize`).

## Frontend

- `authStore.teamLead` signal + `selectLeadsAnyTeam`; a `TeamLeadRoute` guard and a **"My Teams"** nav item visible only to leads (an admin also passes the route).
- New non-admin `useMyTeams` hooks + `teamsApi` client, in a query-key namespace separate from the admin team hooks.
- **My Teams** landing: teams you lead as interactive management cards, member-only teams shown informationally; a detail page reusing the member table / `TeamRoleBadge`, with a member picker driven by the non-admin principal directory. The last-lead guardrail surfaces as an error toast.

## Test plan

- [x] Backend unit (`TeamServiceTest`): lead guard forbids a non-lead, ADMIN bypasses, last-lead / self-lockout guardrails, `listMyTeams`/`leadsAnyTeam`.
- [x] Backend web IT (`MyTeamsControllerIT`, Testcontainers → CI): LEAD-of-X `200`, LEAD-in-Y `403`, MEMBER `403`, ADMIN `200`, `LAST_LEAD` guard, `teamLead` on `/users/me`.
- [x] Frontend: `useMyTeams` hooks call the non-admin client; `TeamLeadRoute` and nav gating visible only to leads; `MyTeamsPage` / `MyTeamDetailPage` member actions + guardrail toast. 830 tests green; coverage gate green (86.7% / 80.9% / 83.3% / 87.5%).
- [x] Local gates: backend compile + ArchUnit + Spotless; frontend lint + format + typecheck.

🤖 Co-Author: Claude (Opus 4.x) via Claude Code


qnop-core/src/main/java/io/qnop/service/TeamService.java
```diff
@@ -181,6 +182,90 @@ public void removeMember(UUID teamId, UUID userId) {
...
+  /**
+   * Blocks an operation that would strip the team's last remaining lead. Only fires when {@code
+   * userId} is currently a LEAD and no other lead remains — removing/demoting a MEMBER, or a LEAD
+   * with co-leads, is unaffected. This is the single place the "never zero leads" invariant lives
+   * for the self-management surface, and it is what prevents a sole lead's self-lockout.
+   */
+  private void guardNotLastLead(UUID teamId, UUID userId) {
+    boolean targetIsLead =
+        memberships.existsByTeamIdAndUserIdAndTeamRole(teamId, userId, TeamRole.LEAD);
+    if (targetIsLead && memberships.countByTeamIdAndTeamRole(teamId, TeamRole.LEAD) <= 1) {
```
devtank42, line 355: **MEDIUM — TOCTOU race can violate the "a team must keep at least one lead" invariant.**

`guardNotLastLead` reads the lead count in its own transaction, but there is no row lock, no unique/check constraint, and `@Version` on `TeamMembership` only guards the *individual* rows being mutated.

Failure scenario: a team has exactly two co-leads `leadA` and `leadB`. `leadA` sends `DELETE /teams/{id}/members/{leadB}` while `leadB` concurrently sends `PATCH /teams/{id}/members/{leadA}` (role → MEMBER). Under Postgres READ COMMITTED each transaction reads the pre-commit count of `2`, both guards pass, both commit, and the team ends up with **zero leads** — the exact invariant this guard and the OpenAPI description promise. Optimistic locking does not catch it because the two transactions touch different membership rows.

Suggested fix: serialize guard + mutation per team (pessimistic `SELECT … FOR UPDATE` on the team's memberships before counting, or a `SERIALIZABLE` tx), and/or add a DB-level backstop constraint. Worth a concurrency test covering the two-request race.

qnop-api/src/main/resources/openapi/openapi.yaml
```diff
@@ -1299,6 +1307,185 @@ paths:
...
+        - bearerAuth: []
+      parameters:
+        - $ref: '#/components/parameters/TeamIdParam'
+      responses:
+        '200':
+          description: The team with its members.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdminTeamDetail'
```
devtank42: **LOW — the non-admin LEAD-facing `/teams/**` responses reuse the admin DTOs (`AdminTeamDetail`, `AdminTeamMember`, `AdminTeamMemberRequest`, `AdminTeamMemberRoleUpdateRequest`).**

Not a defect today, but it couples the non-admin contract to the admin one: if an admin-only field (e.g. audit metadata) is later added to `AdminTeamDetail`/`AdminTeamMember`, it will automatically leak into the non-admin `/teams/{teamId}` response that a plain LEAD can read. Consider a short note here or a follow-up issue to split the schemas when/if that divergence appears.

qnop-app/src/test/java/io/qnop/web/MyTeamsControllerIT.java
```diff
@@ -0,0 +1,310 @@
...
+    // Remove bob.
+    mockMvc
+        .perform(
+            delete("/api/v1/teams/{id}/members/{uid}", teamId, bob.getId())
+                .header("Authorization", bearer(leadToken)))
+        .andExpect(status().isNoContent());
+  }
+
+  @Test
+  void leadCannotManageAnotherTeam() throws Exception {
```
devtank42, line 137: **LOW — cross-team IDOR coverage gap.**

`leadCannotManageAnotherTeam` exercises only `GET` and `POST` against another team. The authz-critical `PATCH`/`DELETE` on another team are covered only at the `TeamService` unit level, not end-to-end through Spring Security + MockMvc. The guard code is identical, so risk is low, but adding the two negative cases here would close the matrix on the most sensitive path.

devtank42:
Addressed all three review findings in 7d95006:

- **MEDIUM — TOCTOU last-lead race:** `setMemberRoleAsLead` / `removeMemberAsLead` now take a pessimistic write lock on the team (`TeamRepository.findByIdForUpdate`, the same `@Lock(PESSIMISTIC_WRITE)` pattern as `DocumentRepository` for #324) *before* the last-lead count, so two concurrent demote/remove requests on different membership rows serialize instead of both committing into zero leads. Added `TeamLeadConcurrencyIT` firing the exact two-request race (co-lead A removes B while B demotes A) and asserting one 2xx, one `409 LAST_LEAD`, and a team that still has exactly one lead.
- **LOW — cross-team IDOR coverage:** `leadCannotManageAnotherTeam` now also exercises `PATCH` and `DELETE` against another team end-to-end (both `403`), closing the matrix on the most sensitive path.
- **LOW — non-admin reuses admin DTOs:** added a note on `getMyTeam` in `openapi.yaml` documenting the deliberate reuse and the divergence risk, and opened #508 to track splitting out non-admin `TeamDetail`/`TeamMember` schemas if an admin-only field is ever added.

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

Issue #464, opened 2026-07-12
Frontend i18n: user-selectable UI language


devtank42:
## Summary

The `qnop-ui` SPA is English-only: every user-facing string is a hardcoded literal and there is no way to switch languages. Introduce a frontend internationalization (i18n) runtime and a **language switcher** so a user can choose their UI language. The chosen language must persist per user and survive reloads.

## Why now

The backend already ships the persistence and defaults for this — only the frontend runtime and the switcher are missing:

- **Per-user preference:** `UserSettingKey.PREFERRED_LANGUAGE` (`preferred_language`, ISO 639-1, default `"en"`), read/written via `GET`/`PUT /users/me/settings` (`getCurrentUserSettings` / `updateCurrentUserSettings`, issue #22).
- **System default:** `ApplicationSettingKey.GENERAL_DEFAULT_LANGUAGE` (`general.default_language`, ISO 639-1, default `"en"`) — the fallback UI language when a user has expressed no preference.
- **Locale helpers already present in the UI:** `qnop-ui/src/utils/locale.ts` (`localeDisplayName`, `localeShortCode`).
- **Precedent:** mail templates already resolve per locale with a fallback chain (requested locale → configured default → `en`, see `MailTemplateService`). The UI should mirror this resolution order so email and UI language stay consistent.

So the data model is done; this issue is the frontend consumption + switcher.

## Goal

A signed-in user can pick their UI language from a switcher; the whole SPA re-renders in that language immediately, the choice is saved to `PREFERRED_LANGUAGE`, and it is restored on next login / reload. When the user has no stored preference, the UI falls back to `general.default_language`, then to the browser language, then to `en`.

## Scope

**In scope**
- Add an i18n runtime to `qnop-ui` (library choice below — likely an ADR).
- Externalize UI strings into translation catalogs, starting with a **pilot surface** (login + app shell + user profile/settings) rather than boiling the ocean in one PR — see "Phasing".
- Ship at least **English (`en`)** and **German (`de`)** catalogs (German is the team's working language; it exercises pluralization/formatting and proves the pipeline).
- A **language switcher** in the app shell (next to the existing theme toggle in `TopBar`) and/or on the profile/settings page.
- Language-resolution order: stored `preferred_language` → `general.default_language` → browser `navigator.language` → `en`.
- Persist the selection to `PREFERRED_LANGUAGE` via the existing user-settings endpoint; apply the value optimistically for instant re-render.
[...]

PR #507, opened 2026-07-17
feat: L10n — UTC storage, ISO-offset API timestamps, user-timezone display (#465)

### LINKED ISSUE #465: L10n: UTC storage, timezone-aware API timestamps, and user-timezone display
## Summary

Audit and fix date/time **localization (L10n)** end-to-end so timestamps are stored in UTC, transported over the API as ISO-8601 **with** offset information, and displayed in the **user's configured timezone**. This complements the language i18n work in #464 (language ≠ timezone; both are needed).

Three requirements to verify and, where needed, correct:

1. The backend never stores a *local* (zone-less) date/time — everything is UTC / timezone-aware.
2. The API always returns date/time values in correct ISO-8601 format **with timezone information**.
3. The frontend always renders timestamps in the **user's profile timezone**, falling back to a **general timezone from the application settings**.

## Current state (from a first pass over the code)

Most of the backend is already in good shape; the main gap is the frontend never applying the stored timezone, plus a missing app-level default.

### 1. Storage — mostly ✅, one thing to confirm
- Every domain timestamp field is `java.time.Instant` (27 imports across `io.qnop.entity`, e.g. `createdAt`, `updatedAt`, `expiresAt`, `dueAt`, `committedAt`). `Instant` is a UTC epoch — zone-agnostic by construction, so no local wall-clock time is persisted.
- 40+ Liquibase columns are `TIMESTAMP WITH TIME ZONE`.
- **To confirm:** `0006-shedlock-schema.yaml` uses plain `TIMESTAMP` (without time zone) for `lock_until` / `locked_at`. That is ShedLock's own schema convention (it works in UTC), so it is very likely fine — but it should be explicitly verified and documented as intentional, not silently different.

### 2. API — likely ✅, needs an explicit guarantee + test
- No explicit Jackson config in `application.yml`, so Spring Boot defaults apply: `JavaTimeModule` + `WRITE_DATES_AS_TIMESTAMPS=false` → `Instant` serializes as ISO-8601 with a `Z` offset (e.g. `2026-07-12T10:15:30Z`).
- `openapi.yaml` declares timestamp fields as `format: date-time` (27 occurrences).
- **To do:** make the guarantee explicit and regression-proof rather than relying on framework defaults (a future dependency bump shouldn't be able to silently flip it).

### 3. Frontend — ❌ the real gap
- `UserSettingKey.TIMEZONE` (`timezone`, IANA id, default `UTC`) already exists and is editable via `GET`/`PUT /users/me/settings` — but it is **never consumed anywhere in `qnop-ui`** (grep finds zero references).
- `qnop-ui/src/utils/formatDate.ts` / `relativeTime.ts` build `Intl.DateTimeFormat('en-GB', { … })` with **no `timeZone` option**, so every timestamp renders in the *browser's* local timezone, ignoring the user's configured preference.
- **There is no general timezone application-setting.** `ApplicationSettingKey` has `GENERAL_DEFAULT_LANGUAGE` but no timezone equivalent — the required fallback source (requirement 3) does not exist yet and must be added.

## Scope

**In scope**
- **Backend audit (requirement 1):** verify no domain code uses `LocalDateTime`/`LocalDate`/`OffsetDateTime`/`java.sql.Timestamp` for stored timestamps, no non-UTC DB defaults (`now()`/`current_timestamp` on a session TZ), and that the JVM/DB timezone can't skew stored values. Document the ShedLock plain-`TIMESTAMP` columns as intentional. Add a lightweight regression guard (ArchUnit rule or a test) that entity timestamp fields are `Instant`.
- **API guarantee (requirement 2):** set `spring.jackson.serialization.write-dates-as-timestamps=false` explicitly (defensive), and add a serialization test asserting a representative endpoint returns an offset-bearing ISO-8601 string (ends in `Z`/`±hh:mm`), never epoch millis and never a zone-less local string.
- **New application setting (requirement 3 fallback):** add `ApplicationSettingKey.GENERAL_DEFAULT_TIMEZONE` (`general.default_timezone`, IANA id, default `"UTC"`), surfaced in the admin Application Settings form, validated as a real IANA zone.
- **Frontend display (requirement 3):** resolve the active timezone as **user profile `timezone` → `general.default_timezone` → `UTC`**, and thread it into all date/time formatters. Centralize in `formatDate.ts` / `relativeTime.ts` so components format through one place; pass `timeZone` to every `Intl.DateTimeFormat`. Show the active zone where a raw timestamp could otherwise be ambiguous (e.g. due dates, audit trails).
- **Tests:** unit tests that the same `Instant` renders differently under different configured zones; that the fallback chain resolves correctly; that changing the profile timezone updates the display.

**Out of scope**
- UI *language* translation — that's #464.
- Letting users pick a timezone *per document/review* — the profile timezone is the single display zone.
- Rewriting timestamps already stored (no data migration needed — they're already UTC).

## Acceptance criteria

- [ ] Confirmed & documented: all persisted timestamps are UTC / zone-aware; ShedLock's plain-`TIMESTAMP` columns verified as intentional; a regression guard prevents zone-less domain timestamp fields.
- [ ] API contractually returns ISO-8601 with an explicit offset for every timestamp; covered by a serialization test; `write-dates-as-timestamps=false` set explicitly.
- [ ] `ApplicationSettingKey.GENERAL_DEFAULT_TIMEZONE` added, validated, and editable in the admin settings UI.
- [ ] Frontend renders every timestamp in the user's profile timezone, falling back to the application default, then `UTC`; no formatter renders in the raw browser zone anymore.
- [ ] Unit tests cover zone-aware rendering and the fallback chain; `pnpm lint`/`format:check`/`typecheck`/`build` and the backend gate are green.

## Notes

- Consider an **ADR** for the timezone policy (store UTC, transport ISO-with-offset, render in user zone) per working rule 5 — it's a cross-cutting decision worth recording.
- Related: language i18n #464; accessibility audit #460.
- Reuse: `UserSettingKey.TIMEZONE`, `GET`/`PUT /users/me/settings`, `qnop-ui/src/utils/formatDate.ts` & `relativeTime.ts`, the `ApplicationSettingsForm` admin surface, and the `Intl`/`date-fns` tooling already present.

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

devtank42:
## Summary

Makes date/time localization correct end-to-end (issue #465, ADR-0039): timestamps are stored in UTC, transported over the API as ISO-8601 **with an explicit offset**, and rendered in the **user's configured timezone** (falling back to an application default, then UTC). Complements the language i18n work in #464.

Most of the backend was already sound (every entity timestamp is `Instant`, columns are `TIMESTAMPTZ`); the real gaps were an implicit/unguarded policy and a frontend that always rendered in the raw browser zone.

### Backend — store UTC, transport ISO-offset (`ced586e`)
- **Store (req 1):** new ArchUnit rule `entityTimestampsAreStoredAsInstant` forbids zone-less/zoned temporal fields in `io.qnop.entity`. ShedLock's plain-`TIMESTAMP` lock table is not a JPA entity and stays exempt (already documented).
- **Transport (req 2):** pinned `spring.jackson.datatype.datetime.WRITE_DATES_AS_TIMESTAMPS: false` and added `TimestampSerializationTest` asserting `Instant` → `...Z`, never epoch millis. _(Jackson 3 moved this feature from `SerializationFeature` to `DateTimeFeature`, bound under `spring.jackson.datatype.datetime` with the exact enum-constant key.)_
- **App default (req 3 fallback):** `ApplicationSettingKey.GENERAL_DEFAULT_TIMEZONE` (IANA id, default `UTC`), validated via new `SettingConstraints.ValueFormat.TIMEZONE` (`ZoneId.of`), seeded, and reported on `/config` as `general.defaultTimezone` (replaces the hardcoded value). Surfaces automatically in the admin Application Settings form.
- **ADR-0039** records the policy.

### Frontend — render in the user's zone (`d98d859`)
- Resolve the active zone as **user profile `timezone` → `general.defaultTimezone` → UTC** (`resolveTimeZone` + `useCurrentUserSettings`/`useDisplayTimezone`).
- `TimezoneSync` (mounted once at the app root, inside the QueryClientProvider) pushes the resolved zone into the UI store, so components format through `useFormatters()` without each depending on a QueryClient.
- `formatDate.ts`/`relativeTime.ts` are now zone-aware (per-zone `Intl` cache, default UTC); **all 11 timestamp call sites** migrated to the `useFormatters()` seam — no formatter renders in the raw browser zone anymore.

## Test plan
- [x] Backend: `spotlessCheck`, `:qnop-core:test`, and qnop-app ArchUnit / ConfigController / serialization tests — green locally.
- [x] Frontend: `tsc`, `eslint`, `prettier --check`, full `vitest` suite (590 tests) — green.
- [x] New tests: fallback-chain resolution, same-instant-different-zone rendering, zone-aware formatter hook, timezone validation, ISO-offset serialization.
- [ ] CI verifies the Docker-only integration tests (no local Docker).

Closes #465.

🤖 Co-Author: Claude (Opus 4.x) via Claude Code


qnop-ui/src/components/reviews/panel/CommentBubble.tsx
```diff
@@ -71,6 +71,7 @@ export function CommentBubble({
   notify,
 }: CommentBubbleProps) {
   const theme = useTheme();
+  const { shortRelativeTime } = useFormatters();
```
devtank42: **HIGH — precise-time tooltip still renders in the raw browser zone** (real target: `const TIME_FORMAT` at line 33 and `title={TIME_FORMAT.format(new Date(createdAt))}` at line 114 — commenting here because those lines are outside this PR's diff hunks).

The visible label was correctly migrated to the zone-aware `useFormatters().shortRelativeTime` (added right here), but the module-level `TIME_FORMAT = new Intl.DateTimeFormat(undefined, …)` used for the hover tooltip has no `timeZone` option, so it defaults to the host/browser zone.

Failure scenario: a reviewer whose configured display zone is `Europe/Berlin` but whose OS/browser is `America/Los_Angeles` sees the relative label computed in Berlin time while the tooltip shows Los-Angeles time — a ~9h-inconsistent tooltip on the same element. This is exactly the class of bug ADR-0039 exists to prevent and contradicts the PR's "no formatter renders in the raw browser zone" / "all 11 call sites migrated" claim.

Fix: render the tooltip through a zone-aware `useFormatters()` full-precision formatter instead of the module-level `Intl.DateTimeFormat`.

qnop-ui/src/components/reviews/panel/AnnotationHead.tsx
```diff
@@ -50,6 +50,7 @@ interface AnnotationHeadProps {
  */
 export function AnnotationHead({ annotation, unseen = false }: AnnotationHeadProps) {
   const theme = useTheme();
+  const { shortRelativeTime } = useFormatters();
```
devtank42: **HIGH — same raw-browser-zone tooltip bug as CommentBubble** (real target: `const DATE_FORMAT` at line 34 and `title={DATE_FORMAT.format(new Date(annotation.createdAt))}` at line 124 — those lines are outside this PR's diff hunks, so anchoring here).

The visible text now uses the zone-aware `useFormatters().shortRelativeTime` (added here), but the tooltip still formats via a module-level `DATE_FORMAT = new Intl.DateTimeFormat(undefined, …)` with no `timeZone`, i.e. the browser zone. When the user's configured zone differs from the host zone, the tooltip disagrees with the label. Fix: route the tooltip through the zone-aware formatter seam.

qnop-core/src/main/java/io/qnop/service/ValueValidator.java
```diff
@@ -101,6 +103,7 @@ private static void validateConstraints(
       switch (constraints.format()) {
         case EMAIL -> requireEmail(keyForError, value);
         case URL -> requireHttpUrl(keyForError, value);
+        case TIMEZONE -> requireTimezone(keyForError, value);
```
devtank42, line 107: **MEDIUM — the per-user `timezone` setting is never format-validated; ADR-0039 overstates what's enforced.**

This PR wires `ValueFormat.TIMEZONE` (`ZoneId.of`) validation onto `ApplicationSettingKey.GENERAL_DEFAULT_TIMEZONE` (great), but the `UserSettingKey` validation path has no equivalent — `UserSettingKey.TIMEZONE` is a plain `STRING`, and `UserSettingsService.updateSettings` persists any string verbatim.

ADR-0039 states "**Both** the per-user and application-default zone values are validated as real IANA ids (`ZoneId.of`) at the setting boundary" — which is not true for the per-user path.

Failure scenario: `PUT` the user-settings endpoint with `{"timezone":"not a real zone"}` → 200 OK, garbage persisted in `user_setting`. The frontend `isValidTimeZone` guard masks it today (falls back to app default/UTC), but any future *backend* consumer of the per-user zone (server-rendered PDF export, scheduled email in the user's zone) gets no such protection.

Fix: give `UserSettingKey` the same `SettingConstraints`/`ValueFormat.TIMEZONE` mechanism and validate it in `UserSettingsService` — or narrow the ADR wording to "application-default only; per-user zone is defensively validated client-side."

devtank42:
Addressed all three review findings (commit `3f2722a`) and merged the latest `main`:

**HIGH — raw-browser-zone tooltips (CommentBubble & AnnotationHead).** Both precise-time hover tooltips now render through the zone-aware `useFormatters().formatDateTime` instead of a module-level `Intl.DateTimeFormat(undefined, …)`, so the tooltip and the visible relative label always agree on the user's display zone. Note: `main` has since replaced `CommentBubble` with `CommentMessage` — I applied the same fix there, and swept the other timestamp surfaces the redesign added (`CommentMessage`, `AnnotationListItem`, `RepliesCard`, `ActivityCard`, `ProfileMovesCard`) onto the `useFormatters()` seam too.

**MEDIUM — per-user timezone not backend-validated.** `UserSettingKey` now carries the same `SettingConstraints` mechanism as `ApplicationSettingKey`, and `UserSettingKey.TIMEZONE` is validated as a real IANA id (`ValueFormat.TIMEZONE` → `ZoneId.of`) in `ValueValidator` — which `UserSettingsService.updateSettings` already invokes. A `PUT /users/me/settings` with a bogus zone is now rejected, so ADR's "both values are validated" is true. Added `enforcesTimezoneFormatForPerUserSetting` to `ValueValidatorTest`.

**Merge + ADR renumber.** Merged `origin/main`; the ADR is renumbered **0039 → 0041** (main landed its own 0039/0040). All gates green locally: `spotlessCheck`, backend `:qnop-core`/arch/config/serialization tests, and the full frontend suite (**817 tests**, incl. the new call sites).

Issue #16, opened 2026-06-14
Application settings service + admin endpoints


bigpuritz:
## Goal
Runtime application-settings service + admin endpoints.

## Scope
- `ApplicationSettingsService`: DB-backed, lock-free `AtomicReference` snapshot, refresh in `afterCommit`, change listeners (e.g. SMTP).
- Key registry (`ApplicationSettingKey`) with defaults + shared `ValueValidator`.
- Secret redaction (`ConfigurationKeyRedactor` / tree): PASSWORD masked as `***`, encrypt on write.
- Endpoints: `GET/PATCH /admin/settings` (superadmin).

Part of #7 · Depends on #10, #13

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

PR #667, opened 2026-08-01
feat(tracking): usage tracking through a server-side proxy, path-anonymised and consent-gated (#666)

### LINKED ISSUE #666: Usage tracking: make the placeholder real — six backends, server-proxied and path-anonymised
## What

`tracking.enabled` and `tracking.provider` have existed since #16 and are read by nothing. This turns the placeholder into a working feature: six supported analytics backends, measured through a server-side proxy, with page views reported as route patterns rather than paths.

## How the data leaves the browser

Through qnop, never past it:

```
Browser ──GET /t/s.js──►  qnop ──►  https://matomo.internal/matomo.js   (cached)
        ──POST /t/e───►  qnop ──►  https://matomo.internal/matomo.php   (forwarded)
```

`PathAwareCspHeaderWriter` pins the SPA to `script-src 'self'; connect-src 'self'` (ADR-0040). Rather than punching a hole in that per deployment, the server fetches the vendor script and forwards the events. Three consequences, all wanted: the CSP stays exactly as strict as it is today, a reviewer's browser never talks to the analytics host, and ad blockers have nothing to block.

The forwarded client IP is anonymised by default (last octet / lower 64 bits zeroed) — without any IP the tracker counts every reviewer as one visitor, and with the full one the whole point of proxying is lost.

The configured host goes through the existing SSRF policy (`OidcSsrfPolicy`, #21). Self-hosted trackers usually live on private addresses, so the exception is a deployment property (`qnop.tracking.allow-private-host`), never a database setting an admin could flip for themselves.

## Providers

Matomo, Plausible, Umami, PostHog (self-hostable) plus Fathom and Pirsch (cookieless cloud). PostHog is pinned shut: no autocapture, no session recording, no automatic page views — autocapture collects click text, and in qnop click text contains document titles.

**Explicitly out of scope, permanently:** Hotjar, Clarity, FullStory, Smartlook and everything else that records sessions or draws heatmaps. They capture the screen, and the screen holds a customer's confidential document.

## What is measured

Page views as **route patterns**, derived from the router's own params:

```
/reviews/8f3c…-a1/tasks?filter=open   →   /reviews/:documentId/tasks
```

Plus four product events with no properties at all: `review_created`, `annotation_created`, `review_finalized`, `export_generated`. What is counted is that something happened — not to which document.

## Opting out

Four independent gates; all four must agree before anything loads:

- **DNT/GPC** honoured (`tracking.respect_dnt`, on by default)
- **Consent** asked for before the first load (`tracking.consent_required`), revocable in the profile
- **A personal switch** in the profile (`privacy.usage_tracking_opt_out`), stored per account so it survives a device change
- **Privileged roles** (ADMIN, AUDITOR) measured only when `tracking.track_privileged_roles` is explicitly on — off by default, since a handful of admins is barely statistics but very much personal data

## Not in scope

- Server-side tracking of API calls
- Custom/arbitrary analytics scripts
- Scheduling, funnels, cohort analysis — whatever the chosen backend does, it does on its own side

bigpuritz:
Closes #666.

Turns the `tracking.*` placeholder from #16 into a working feature: five analytics backends, measured through a server-side proxy, with page addresses reported as route patterns rather than URLs.

## Why a proxy

`PathAwareCspHeaderWriter` pins the SPA to `script-src 'self'; connect-src 'self'` (ADR-0040), and every analytics vendor's install instructions begin by violating exactly that. Extending the CSP per deployment was the obvious move and was rejected: it weakens the one policy that is identical on every other page, hands each reviewer's IP to a third party, and puts the measurement in an ad blocker's path.

```
Browser ──GET /t/s.js──►  qnop ──►  https://analytics.example/script.js   (cached 1h)
        ──POST /t/c/…─►  qnop ──►  https://analytics.example/api/send     (forwarded)
```

The CSP stays untouched, the browser never contacts the analytics host, and the forwarded address is truncated to its /24 or /64 — without any, the backend counts a whole company as one visitor; with the real one, proxying has withheld nothing.

Two boundaries are enforced by the server rather than trusted to the client:

- **The collect allowlist.** Only the paths a provider declares are forwarded, matched whole. PostHog's `/s/` is absent, which makes "qnop does not record sessions" a property of this server instead of a client setting somebody could change.
- **The URL sanitiser.** Every forwarded measurement loses its query string and its identifier-shaped segments, so a stale bundle in an open tab cannot put a document id into a report that outlives the review.

## Backends

Matomo, Plausible, Umami, PostHog, Pirsch. **Fathom was dropped during implementation** — its custom-domain feature was withdrawn in 2023 with no replacement, and its model expects beacons to arrive directly from the browser so its bot detection sees the real client. Proxying it would quietly poison its data, and a CSP exception for one vendor would give up the property this design exists to keep.

Session recording and heatmaps (Hotjar, Clarity, FullStory, …) are out permanently: they record the screen, and in qnop the screen holds a customer's document.

## What is measured

`/reviews/:documentId/tasks`, built from the router's own params — the only way to catch a slug like `/users/mia-member` that no shape check recognises as personal. Plus four events with no properties: `review_created`, `annotation_created`, `review_finalized`, `export_generated`.

Four gates, any of which says no on its own: operator configuration, DNT/GPC, the account-level opt-out (in the profile, so it follows a person across devices), and consent. Administrators and auditors are excluded unless explicitly included.

## Found by running it

The reporter configured Umami against a real instance and saw nothing. Three fixes came out of that, and they are the most valuable part of this branch:

1. **`/t` was not proxied by the dev server.** It sits outside `/api` on purpose, and therefore outside the one Vite proxy rule. In production the backend serves the SPA (ADR-0040), so it worked in a deployment and failed on every developer machine.
2. **A complete-looking configuration measured nothing** because the master toggle was still off, and nothing on the screen said so. The settings page now states whether measurement is running and, when it is not, which of the four conditions is missing.
3. **Page titles were travelling.** Every one of these backends sends `document.title` beside the address (Umami `title`, Matomo `action_name`, PostHog `$title`). qnop sets no per-page title today, so nothing leaked — the first person to make browser tabs say what they show would have started exporting review titles. A title has no structure to anonymise, so it is dropped server-side and never set client-side.

The proxy also stopped swallowing backend failures: one WARN when forwarding starts failing, one INFO when it recovers, on the transition rather than per measurement.

Also here: the OIDC SSRF guard is extracted into `OutboundUriGuard` and shared, rather than the check being copied for a second operator-supplied fetch target; and the Testcontainers Postgres gets `max_connections=300`, because each test class with its own `@TestPropertySource` holds its own pool and the default ceiling was one context away from failing an unrelated class.

## Test plan

- [x] `./gradlew build` — green (Spotless, ArchUnit, full suite incl. Testcontainers)
- [x] `pnpm typecheck`, `eslint`, `vitest` — green, 1265 tests
- [x] `TrackingForwardIT` drives the whole path against a stand-in backend on loopback: the script is served from this origin, and a measurement arrives with the document id, the query string and the title removed; an undeclared path never leaves
- [x] Unit: provider allowlist (incl. session recording refused), URL sanitiser, IP truncation, config resolution with its SSRF cases, payload rewriting, the four gates, path anonymisation, the settings status line
- [x] **Manual, against a real Umami**: script loads (200), page view leaves as `{"payload":{"website":"…","url":"/login","title":"","referrer":""}}`, and Umami accepts such events (200 + session token)
- [ ] Worth doing before merge: the same manual check against one more backend — the other four integrations are built from vendor documentation, not from a running instance

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ qnop-ui/.vite/
 /*.xlsx
 /*.pdf
 /*.csv
+.playwright-mcp/
--- a/docs/adr/0056-usage-tracking.md
+++ b/docs/adr/0056-usage-tracking.md
@@ -1,0 +1,69 @@
+# ADR-0056: Usage tracking — server-proxied, path-anonymised, consent-gated
+
+- **Status:** Accepted
+- **Date:** 2026-08-01
+- **Deciders:** devtank42 (with Claude)
+
+## Context
+
+`tracking.enabled` and `tracking.provider` have existed since issue #16 and were read by nothing — a dropdown offering Matomo, Plausible and Umami, wired to no code at all. Making it real means answering questions a document-review tool cannot answer the way an ordinary web app would.
+
+The binding constraint is already in the repo. `PathAwareCspHeaderWriter` (ADR-0040) pins the SPA to `script-src 'self'; connect-src 'self'`. Every analytics vendor's install instructions begin by violating exactly that.
+
+And the second constraint is the product itself: a qnop URL contains the id of a customer's document, a search query is the subject of one, and the visible screen *is* one.
+
+## Decision
+
+### Measurement goes through this server, never past it
+
+```
+Browser ──GET /t/s.js──►  qnop ──►  https://matomo.internal/matomo.js   (cached 1h)
+        ──POST /t/c/…─►  qnop ──►  https://matomo.internal/matomo.php   (forwarded)
+```
+
+The alternative — extending the CSP with the configured analytics host per deployment — was rejected. It weakens the one policy that is identical on every other page, it hands each reviewer's IP address to a third party, and it puts the measurement squarely in the path of ad blockers. Proxying costs two endpoints and turns all three around: the CSP stays untouched, the backend sees only what qnop chooses to send, and there is no third-party host to block.
+
+The endpoints sit outside `/api/v1` (as `@Controller`, since `ApiPathConfig` prefixes every `@RestController`): they speak whatever shape the configured backend speaks, which is that backend's contract and not qnop's published one (ADR-0015).
+
+### The collect allowlist is a boundary, not a convenience
+
+`TrackingProvider` declares, per backend, the exact paths that may be forwarded — compared whole, never by prefix. This is what makes a promise enforceable rather than merely stated: PostHog's `/s/` (session recording) is absent, so a client that asked for session replay would be refused *by this server*, whatever its configuration says. It also stops the endpoint from being a general-purpose relay to the analytics host.
+
+**Session recording and heatmaps are out of scope permanently** — Hotjar, Clarity, FullStory, Smartlook and everything like them. They record the screen, and in qnop the screen holds a customer's confidential document. That is not a setting.
+
+### Addresses are patterns, and they are checked twice
+
+The client reports `/reviews/:documentId/tasks`, built from the router's own params — the router knows which segments were parameters because it filled them in, which is the only way to catch a slug like `/users/mia-member` that no shape check could recognise as personal.
+
+The server checks again on every forwarded measurement (`TrackedUrlSanitizer`): the query string always goes, and identifier-shaped segments become `:id`. A stale bundle in an open tab, or a page that forgets the rule, cannot put a document id into an analytics report that will outlive the review.
+
+### The visitor's address is truncated, not withheld
+
+Proxying raises a question direct measurement never asks. Forward nothing and the backend counts a whole company as one visitor; forward the real address and proxying has withheld nothing. So `X-Forwarded-For` carries the address truncated to its /24 (IPv4) or /64 (IPv6) — Matomo's own anonymisation, and the long-standing middle ground of German data-protection practice. `tracking.forward_client_ip=none` remains available, with distorted counts as the stated cost.
+
+### Four gates, any of which says no
+
+Operator configuration, Do-Not-Track/GPC, the account-level opt-out, and consent. All four must agree before a script tag is even created. The consent answer lives in the browser (the sign-in screen is measured, and nobody has an account yet at that point); the opt-out lives on the account, so it follows a person across devices and outlives a cleared browser store. Administrators and auditors are excluded unless an operator explicitly includes them — a handful of privileged people is barely a statistic and very much personal data.
+
+### Events are a closed union
+
+Four names, no properties: `review_created`, `annotation_created`, `review_finalized`, `export_generated`. An `event(name, props)` API would eventually carry a document title or a search term — not through carelessness, but because that is what an open API invites. A fifth event is a code change somebody reviews.
+
+### Which backends, and one that did not fit
+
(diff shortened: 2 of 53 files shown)


bigpuritz:
Verified against real Matomo and Plausible instances in Docker, both through the proxy.

**Matomo 5.12** (self-hosted, `idSite=1`) — the hit as stored in `matomo_log_action`:

| url | page_title |
|---|---|
| `localhost:5173/reviews/:id/compare` | `NULL` |

**Plausible CE v3.0.1** (self-hosted, domain `qnop.local`) — as stored in ClickHouse `events_v2`:

| pathname | name | hostname |
|---|---|---|
| `/reviews/:id/tasks` | `pageview` | `qnop.local` |

In both: document id replaced, query string gone, page title removed.

This turned up a real bug, now fixed in 2103b7d0: **Matomo and Pirsch measure with GET**, the other three POST, and the proxy accepted POST only — so `GET /t/c/…` answered 405 and two of five providers could never have worked. It stayed hidden because Umami, the first backend this was tried against, posts. Two smaller versions of the same mistake came with it: the rate limiter counted POSTs only, leaving the GET-based providers unbounded on an unauthenticated endpoint, and the collect path was read straight off the request URI, so a query left on it failed the allowlist check silently.

Coverage for the remaining two (PostHog, Pirsch) is still vendor-documentation only.

bigpuritz:
Checked PostHog and Pirsch as well — against their real hosts, since neither can be run locally the way Matomo and Plausible were (PostHog's self-hosted stack is 34 services; Pirsch requires an enterprise licence to self-host at all).

**Pirsch was broken.** `pa.js` defaults to `/hit`, `/event` and `/session`; the `/pv`, `/e`, `/s` this had assumed answer **404** on api.pirsch.io. Fixed in 23ac186e — provider table and data attributes now use the script's own names.

**PostHog's paths were right**, but its SDK also fetches a remote config (`/array/<token>/config.js`), feature flags and lazily-loaded modules — none on the allowlist, and none that should be. The client now disables all three, so it only ever posts measurements.

Verified through the proxy:

| | script | measurement |
|---|---|---|
| PostHog | `/t/s.js` → 235959 bytes of `array.js` from **us-assets**.i.posthog.com | `POST /t/c/e/` → **us**.i.posthog.com answered **400** |
| Pirsch | `/t/s.js` → 11842 bytes of `pa.js` | `GET /t/c/hit` and `POST /t/c/event` → **400**, identical to calling the host directly |

400 rather than 404 is the point: the requests land on endpoints that exist and are rejected only because the probe credentials are fake. PostHog's two-host split (assets ≠ ingestion) is exercised by the script fetch.

**Still unverified for these two:** ingestion itself. PostHog needs a project token, Pirsch needs a paid account or an enterprise licence. Matomo, Plausible and Umami are confirmed all the way to the stored row.

Running tally of what testing against real backends has found: the GET/POST split (Matomo, Pirsch), the wrong Pirsch endpoints, PostHog's extra fetches, the missing dev-server proxy route, the silent master switch, and page titles travelling. None of these was visible from the vendor documentation alone.

PR #777, opened 2026-08-29
feat(ui): Playwright regression net for responsive layouts

bigpuritz:
## Summary

Stacked on #724's branch (its touch-target check needs that fix); GitHub retargets to `main` once #724 merges.

The #461 audit named a regression net as the missing piece and the project had no E2E framework. This adds Playwright and wires it into the CI **smoke job**, where a built, seeded stack with an uploaded PDF review already exists. The suite drives the working-tree SPA through the Vite dev server, API proxied to that stack via the new `QNOP_API_URL` knob.

- `e2e/responsive.spec.ts` — 16 surfaces + login at all six audit widths: nothing scrolls sideways (reads `documentElement`, `body` **and** the shell's `main`, because below `md` that container swallows exactly the overflow the check exists for, see #723), plus the three header controls measured against 44 px overlay included.
- `e2e/visual.spec.ts` — screenshot baselines of dashboard, reviews list, workspace, tasks board and login at 320/768/1024/1440 (light theme; the theme moves colours, not boxes), 0.5 % pixel tolerance.
- Fixtures: one signed-in context per worker (the refresh cookie rotates, ADR-0026 — a stored state replayed into fresh contexts works exactly once); the smoke review resolved once per run in `globalSetup` (login is rate-limited per IP, ADR-0027); `settle()` instead of `networkidle` (polling queries never go idle) waiting for fonts, no indeterminate indicators/skeletons, a stable `main` scroll size and the PDF canvas on workspace routes; `localStorage` cleared per page so no test inherits another's device-local layout.
- `docs/RESPONSIVE.md` gains findings 5–8, the net and how baselines are updated; README lists `pnpm test:e2e`.

### What the net found on its first run

Each filed and marked `test.fail()` until fixed, so the marker flips on the fix:

- #772 viewer toolbar's trailing group (412 px) does not wrap → workspace overflows at 320/375
- #773 KPI card rows do not wrap (dashboard, storage consistency)
- #774 review head's action row (784 px) overflows beside an open sidebar at 1024 on every review route
- #775 review title collapses to one letter at 1440 (visible in the baselines; not an overflow, so no marker)

Closes #725 · part of #461

## Test plan

- [x] Locally against the smoke stack: responsive 108/108, visual 20/20 in two consecutive runs (baselines stable)
- [x] `pnpm test:run` 1319/1319, `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm audit`
[...]

Issue #725, opened 2026-08-05
Playwright visual-regression net for responsive layouts


bigpuritz:
#461 asks for Playwright screenshot tests at 320/768/1024/1440 wired into CI. The project has **no E2E framework at all** today — `package.json` carries Vitest and `vitest-axe`, nothing else — so this is infrastructure work rather than responsive work, and it is the reason it is split out.

### Scope

- Playwright with a fixture that signs in against seeded data
- Screenshot baselines for the key surfaces at the four widths, both themes where layout differs
- An overflow assertion (`document.documentElement.scrollWidth <= clientWidth`) as a cheap, non-visual guard that catches most regressions without baseline churn
- CI wiring, including how baselines are updated and reviewed

The overflow assertion is worth having even if the screenshot suite is deferred: it is the check that would have caught anything structural, and it needs no images.

From the #461 audit; findings in `docs/RESPONSIVE.md`.

🤖 Co-Author: Claude (Opus 5) via Claude Code

bigpuritz:
One consequence of the #723 fix that this net has to account for: the app shell's content container has `overflow: auto` below `md`, which **masks body-level overflow** — the audit's clean "no body overflow" numbers are partly because that container swallows the scroll. The overflow assertion must therefore check the shell content element (`main`'s scroll container) as well as `document.documentElement`, or it will pass while a surface silently scrolls sideways.

Also: the review page header at 768 px is the first screenshot candidate — its stacking behaviour below `md` is exactly the kind of layout that jsdom cannot verify and this net exists for.

🤖 Co-Author: Claude (Fable 5) via Claude Code

PR #783, opened 2026-08-29
fix(ui): KPI tiles share the row on phones instead of overflowing

### LINKED ISSUE #773: KPI card rows overflow the viewport at 320 px (dashboard, storage consistency)
Found by the responsive regression net (#725): KPI cards laid out as one non-wrapping row run past the viewport on phones.

- `/admin/storage-consistency` at **320 px**: `main` scrolls to **348 px**. The cards ("Storage objects", "Orphaned objects", …) are 166 px each; the second spans x = 182…348.
- Dashboard (`/`) at **320 px**: `main` scrolls to **415 px**. The cards ("Open reviews", "Waiting on you", …) are 169 px each; the second spans x = 174…343. At 375 px the row fits the smoke data by a few pixels — a longer label or a two-digit count tips it over.

Both are the same component pattern, so one fix should cover both — and any other KPI row that uses it.

### Fix

Let the row wrap (or stack the cards at `xs`) so each card takes the full width on a phone. Then flip the `test.fail()` markers in `qnop-ui/e2e/responsive.spec.ts` that reference this issue — the net fails on an unexpected pass, so they cannot be forgotten.

Part of #461.

bigpuritz:
## Summary

At 320 px the KPI rows overflowed `main` (dashboard → 415 px, storage consistency → 348 px; #773, found by the responsive net #725). Both surfaces render the same `StatStrip`, so the fix is one component:

- Grid columns are `minmax(0, 1fr)` instead of `1fr` — a plain `1fr` column never shrinks below its content's min width, and the `noWrap` label set that floor.
- The label may wrap to a second line now instead of truncating ("Resolved this week" stays fully readable on a phone); tiles in the same grid row keep equal height.
- Covers the dashboard, storage consistency and scheduler KPI rows in one change.

**Note on the `test.fail()` markers:** the responsive spec lives in the still-open PR #777; its markers for #773 flip when that branch rebases onto this fix — the net fails on an unexpected pass, so it cannot be forgotten.

## Test plan

- [ ] Pixel verification through PR #777's 320/375 px overflow checks once both meet

Closes #773

### DIFF
--- a/qnop-ui/src/components/dashboard/StatStrip.tsx
+++ b/qnop-ui/src/components/dashboard/StatStrip.tsx
@@ -109,6 +109,10 @@ export function StatStrip({ tiles }: { tiles: StatTile[] }) {
                     lineHeight: 1.2,
                     fontVariantNumeric: 'tabular-nums',
                     color: tile.value > 0 ? color : 'text.primary',
+                    // Defensive: a many-digit value is one unbreakable token;
+                    // without this it would re-open the overflow the
+                    // minmax(0, 1fr) columns just closed (issue #773).
+                    overflowWrap: 'anywhere',
                   }}
                 >
                   {tile.value}
--- a/qnop-ui/src/components/dashboard/StatStrip.tsx
+++ b/qnop-ui/src/components/dashboard/StatStrip.tsx
@@ -56,7 +56,14 @@ export function StatStrip({ tiles }: { tiles: StatTile[] }) {
     <Box
       sx={{
         display: 'grid',
-        gridTemplateColumns: { xs: '1fr 1fr', sm: 'repeat(4, 1fr)' },
+        // minmax(0, 1fr): a plain 1fr column refuses to shrink below its
+        // content's min width, so the longest label pushed the whole grid
+        // past a 320 px viewport (issue #773). With the floor at 0 the tiles
+        // share the row and the labels wrap instead.
+        gridTemplateColumns: {
+          xs: 'repeat(2, minmax(0, 1fr))',
+          sm: 'repeat(4, minmax(0, 1fr))',
+        },
         gap: 1.5,
       }}
     >
@@ -106,7 +113,10 @@ export function StatStrip({ tiles }: { tiles: StatTile[] }) {
                 >
                   {tile.value}
                 </Typography>
-                <Typography variant="caption" color="text.secondary" noWrap component="p">
+                {/* Wrapping beats truncation for a KPI label: on a 320 px
+                    phone "Resolved this week" takes a second line instead of
+                    an ellipsis. */}
+                <Typography variant="caption" color="text.secondary" component="p">
                   {tile.label}
                 </Typography>
               </Box>


bigpuritz:
## Code Review — PR #783 (fix(ui): KPI tiles share the row on phones instead of overflowing)

Reviewed the diff against issue #773, the three consumers (`HomePage.tsx`, `StorageConsistencyPage.tsx`, `SchedulerPage.tsx`), and searched for existing tests/screenshots that depend on `StatStrip`'s previous `noWrap`/ellipsis behavior.

### Findings

**[MEDIUM] Numeric `tile.value` has no overflow safeguard once the column floor is removed**
`qnop-ui/src/components/dashboard/StatStrip.tsx` (the `fontSize: '1.4rem'` value `Typography`)
Before this change, the grid column's implicit `min-content` floor was set by whichever child was widest — in practice the `noWrap` label, but it also happened to cover the value. `minmax(0, 1fr)` removes that floor for the whole column, including for the value text, which still has no `noWrap`/`whiteSpace` handling and is a single unbreakable token (a number has no space to wrap on). All current call sites (`reviews.length`, `orphanedCount`, `jobs.length`, etc.) render small integers today, so this won't reproduce in practice right now, but a KPI that legitimately grows into 5+ digits (e.g. `storageObjectCount` on a large deployment) could silently overflow/clip the tile at 320 px with no fallback — the exact class of bug #773 was filed to close, just shifted from the label to the value. Consider adding `overflowWrap: 'anywhere'` (or a min font-size clamp) to the value `Typography` as a defensive measure, or confirm the value is intentionally out of scope here.

**[LOW] No regression coverage lands in this PR**
The PR body correctly notes the Playwright pixel checks live in the still-open PR #777 and will flip once #777 rebases onto this fix — I confirmed #777 exists and is the actual home of `responsive.spec.ts` (no `e2e/responsive.spec.ts` exists on `main` or this branch yet), so this isn't a silent gap, just flagging that merge order matters: this PR alone doesn't add any automated check (unit or e2e) that would catch a future 320 px regression in `StatStrip` on its own.

### Verified as correct
- `minmax(0, 1fr)` is the right fix for the actual root cause: grid `1fr` tracks default to `minmax(auto, 1fr)`, so a `white-space: nowrap` label set an unshrinkable `min-content` floor on the column — this is a well-documented CSS grid gotcha, and the fix addresses it directly rather than papering over it (e.g. with `overflow: hidden` on the row).
- Dropping `noWrap` on the label in favor of wrapping is reasonable and matches the tone of the code comment; all current labels across the three consumers ("Resolved this week", "Missing binaries", "Storage objects", "Orphaned objects", etc.) are short enough that word-wrapping (not mid-word breaking) is the expected outcome — no consumer passes labels that would stress this further.
- Grid rows still get equal tile heights "for free" via the CSS Grid default `align-items: stretch`, so a wrapped 2-line label in one tile correctly stretches its row-sibling to match — no explicit height logic needed, and none was added.
- Checked all three consumers (`HomePage.tsx`, `StorageConsistencyPage.tsx`, `SchedulerPage.tsx`); all pass 4 tiles each, so the `xs: repeat(2, ...)` grid always fills complete 2-tile rows — no orphaned single-tile last row to worry about.
- No existing unit test, snapshot, or screenshot baseline references `StatStrip`, `noWrap`, or the KPI ellipsis behavior — nothing in this repo's test suite depends on the old truncation behavior.

## Review Summary

| Severity | Count | Status |
|----------|-------|--------|
| CRITICAL | 0     | pass   |
| HIGH     | 0     | pass   |
| MEDIUM   | 1     | info   |
| LOW      | 1     | note   |

Verdict: APPROVE — minimal, well-targeted fix for the reported overflow with correct CSS grid reasoning. The one MEDIUM finding (unbounded numeric value with no wrap fallback) is a pre-existing latent risk this PR incidentally widens slightly, not a new regression caused by the change; worth a follow-up rather than blocking this fix.

bigpuritz:
Review addressed:

- **MEDIUM (unbreakable value token)** → the value Typography now carries `overflowWrap: 'anywhere'`, so a KPI growing past four digits breaks inside the tile instead of re-opening the overflow the `minmax(0, 1fr)` columns just closed.
- **LOW (no automated pixel check in this PR)** → acknowledged; the Playwright assertions live in PR #777 and take over once both meet.

Issue #723, opened 2026-08-05
Review action bar hides "New version" behind an invisible scroll at 768px


bigpuritz:
Measured on the review workspace at a 768 px viewport: the **New version** control spans x = 692…798, so 30 px sit past the edge.

It is not unreachable — the action bar has `overflow-x: auto` and scrolls (`scrollWidth` 798 vs `clientWidth` 768) — but nothing signals that. On a portrait tablet the button reads as clipped, and a primary action ends up behind a gesture nobody is invited to make.

### Options

- Let the bar wrap onto a second line below ~800 px, which costs vertical space but hides nothing
- Keep one row and move overflowing actions into a "more" menu, which keeps the height and needs a priority order
- Keep the scroll and make it visible (edge fade, or a scroll affordance)

Wrapping is probably right here: there are few enough actions that a second row is cheaper than a menu nobody opens.

From the #461 audit; findings in `docs/RESPONSIVE.md`.

🤖 Co-Author: Claude (Opus 5) via Claude Code

PR #784, opened 2026-08-29
fix(ui): the page header stacks when the action row stops fitting

### LINKED ISSUE #774: Review head action bar overflows the content at 1024 px beside an open sidebar
Found by the responsive regression net (#725): on every review route — the workspace (split and focus mode), `/tasks` and `/compare` — at **1024 × 768** the `main` container (764 px wide beside the open sidebar) scrolls to **824 px**. The review hub head's action row — owner, counters, due date, "Add reviewer", "New version" — measures 784 px; "New version" spans x = 978…1084 against a content edge at 1064.

This is #723's sibling one breakpoint up: that fix stacks the header below `md`, but at 1024 px the header is side-by-side again, `PageHeader`'s action slot is `flexShrink: 0`, and with the sidebar open the content is only 764 px wide — narrower than the action row.

### Fix

Either let the action slot wrap at `md` when the content is narrower than the row (stack until the row fits, not until a fixed breakpoint), or raise the stacking breakpoint to `lg` when the sidebar is open. Then flip the `test.fail()` marker in `qnop-ui/e2e/responsive.spec.ts` that references this issue — the net fails on an unexpected pass, so it cannot be forgotten.

Part of #461.

bigpuritz:
## Summary

On every review route at 1024 × 768 beside the open sidebar, `main` (764 px) scrolled to 824 px: #723 stacks the header below `md`, but at 1024 px the header is side-by-side again and the `flexShrink: 0` action slot (784 px review action row) cannot be narrowed (#774, found by the responsive net #725).

The fix makes stacking intrinsic instead of breakpoint-bound, in `PageHeader` (one change for every surface using the canonical header):

- The `md`+ header row is a **wrapping** flex row (`flexWrap` + `useFlexGap`): the action slot drops below the title exactly when it stops fitting — "stack until the row fits, not until a fixed breakpoint", the first option the issue proposes.
- The action slot keeps `flexShrink: 0` (titles still cannot crush buttons) but gains `maxWidth: '100%'`, so once wrapped onto its own line a wide action bar's internal `flexWrap` (already present in `ReviewHubHead`) finally has an edge to break at.
- Below `md` nothing changes (#723's stacking stays).

**Note on the `test.fail()` marker:** the responsive spec lives in the still-open PR #777; its marker for #774 flips when that branch rebases onto this fix.

## Test plan

- [ ] Pixel verification through PR #777's 1024 px overflow checks once both meet

Closes #774

### DIFF
--- a/qnop-ui/src/components/admin/layout/PageHeader.tsx
+++ b/qnop-ui/src/components/admin/layout/PageHeader.tsx
@@ -61,9 +61,15 @@ export function PageHeader({
       // flexWrap never fired because its parent never imposed a limit.
       // Stacking hands the actions the full content width instead, and the
       // wrap that was always there finally has an edge to wrap at.
+      //
+      // Above md the row itself wraps (issue #774): beside an open sidebar
+      // the content can be narrower than the action row even at 1024 px, so
+      // the actions drop below the title exactly when they stop fitting —
+      // stacking follows the available width, not a fixed breakpoint.
       direction={{ xs: 'column', md: 'row' }}
       spacing={2}
-      sx={{ justifyContent: 'space-between', alignItems: { md: 'center' } }}
+      useFlexGap
+      sx={{ justifyContent: 'space-between', alignItems: { md: 'center' }, flexWrap: 'wrap' }}
     >
       <Stack direction="row" spacing={1.5} sx={{ minWidth: 0, alignItems: 'center' }}>
         {leading && <Box sx={{ flexShrink: 0 }}>{leading}</Box>}
@@ -81,7 +87,9 @@ export function PageHeader({
           )}
         </Box>
       </Stack>
-      {action && <Box sx={{ flexShrink: 0 }}>{action}</Box>}
+      {/* maxWidth caps the slot at the line it wrapped onto, so a wide action
+          bar's own flexWrap finally has an edge to break at (issue #774). */}
+      {action && <Box sx={{ flexShrink: 0, maxWidth: '100%' }}>{action}</Box>}
     </Stack>
   );
 }
--- a/qnop-ui/src/components/admin/layout/PageHeader.tsx
+++ b/qnop-ui/src/components/admin/layout/PageHeader.tsx
@@ -88,8 +88,10 @@ export function PageHeader({
         </Box>
       </Stack>
       {/* maxWidth caps the slot at the line it wrapped onto, so a wide action
-          bar's own flexWrap finally has an edge to break at (issue #774). */}
-      {action && <Box sx={{ flexShrink: 0, maxWidth: '100%' }}>{action}</Box>}
+          bar's own flexWrap finally has an edge to break at; ml: 'auto' keeps
+          the slot right-aligned even when it is alone on a wrapped line, where
+          space-between would park a lone item at line start (issue #774). */}
+      {action && <Box sx={{ flexShrink: 0, maxWidth: '100%', ml: { md: 'auto' } }}>{action}</Box>}
     </Stack>
   );
 }
--- a/qnop-ui/src/components/reviews/hub/ReviewHubHead.tsx
+++ b/qnop-ui/src/components/reviews/hub/ReviewHubHead.tsx
@@ -185,7 +185,10 @@ export function ReviewHubHead({
   const shown = participants.slice(0, MAX_STACK_AVATARS);
 
   return (
-    <Stack direction="row" spacing={1.5} sx={{ alignItems: 'center', flexWrap: 'wrap' }}>
+    // useFlexGap: with margin-based spacing a wrapped second row would sit
+    // flush against the first — the wrap fires for real now that PageHeader
+    // hands this bar a width limit (issue #774).
+    <Stack direction="row" spacing={1.5} useFlexGap sx={{ alignItems: 'center', flexWrap: 'wrap' }}>
       {/* The owner is structurally public (issue #472), so the hover card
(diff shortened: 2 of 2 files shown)


bigpuritz:
## Code review — PR #784

Reviewed `PageHeader.tsx` in context of #723, `ReviewHubHead.tsx` (the wide action bar with its own `flexWrap`), and every other `<PageHeader>` consumer (grepped across `qnop-ui/src`).

The core fix is sound: `flexWrap: 'wrap'` + `useFlexGap` on the `md`+ row, plus `maxWidth: '100%'` on the action slot, is the right mechanism — `max-width` is respected by the flex sizing algorithm even with `flexShrink: 0`, so the action slot is finally clamped to the line width instead of overflowing, and `useFlexGap` is exactly what MUI's own docs recommend once `flexWrap` is combined with `Stack`'s `spacing` prop (margin-based spacing doesn't collapse correctly across wrapped lines). No regression found in the plain single-button consumers (`HomePage`, `UsersPage`, `TeamsPage`, `OidcProvidersPage`, `MessagesPage`, `AuditPage`, `StorageConsistencyPage`) — their action content is far narrower than the review head's, so they simply keep the current side-by-side layout at `md`+.

Two things worth fixing before/along with merge:

**[HIGH] The wrapped action line will left-align, not right-align, once it drops to its own row**
`qnop-ui/src/components/admin/layout/PageHeader.tsx` (the `sx={{ justifyContent: 'space-between', ... }}` on the outer `Stack`)

When the action slot still fits next to the title, `justify-content: space-between` on a single flex line pushes it to the far right — correct, that's the current side-by-side look. But once the row actually wraps (the scenario this PR exists to handle), the title and the action end up as the *sole* item on their own respective flex lines. Per the flexbox spec, `justify-content: space-between` with exactly one item on a line has nothing to distribute space between, so the browser places that item at the line's main-start (left) — not main-end (right). Concretely: at 1024 px beside an open sidebar, once "Add reviewer" / "New version" etc. wrap below the title, the whole button row will hug the left edge instead of staying right-aligned, which reads as an unintentional shift rather than a deliberate stacked layout (compare to the `xs` stacked case, where `alignItems` is unset and children stretch full-width — different mechanism, coincidentally also left-looking, but not the same reason). Worth an explicit `alignSelf: 'flex-end'` (or `justifyContent: 'flex-end'` on the action `Box` itself) for the wrapped case, or a comment saying the left shift is accepted.

**[HIGH] `ReviewHubHead`'s own wrapped rows will have zero vertical gap between lines**
`qnop-ui/src/components/reviews/hub/ReviewHubHead.tsx` (`<Stack direction="row" spacing={1.5} sx={{ alignItems: 'center', flexWrap: 'wrap' }}>`)

This Stack was already `flexWrap: 'wrap'`, but per the issue text it "never fired because its parent never imposed a limit" — this PR is precisely what gives it that limit for the first time. It uses `spacing={1.5}` without `useFlexGap`, so MUI applies plain `margin-left` between children instead of CSS `gap`. That only adds horizontal spacing within a line; it does not add vertical spacing between wrapped lines. The first time this component's `flexWrap` actually engages (e.g. at 1024 px beside an open sidebar, or any narrower `main`), the second row of buttons will sit flush against the first row with no vertical breathing room. Since PR #784 is the change that makes this wrap reachable in practice, it should add `useFlexGap` to this Stack in the same PR — otherwise #774 trades a horizontal overflow bug for a visually cramped wrap the first time anyone actually hits it, and there's nothing in `#777`'s described pixel-overflow checks that would catch a spacing/cosmetic issue like this.

**[MEDIUM] No test locks in the new wrap threshold in this PR**
The PR relies entirely on the still-open PR #777's e2e `responsive.spec.ts` checks for verification (per the PR's own test plan and the "Note on the `test.fail()` marker"). That's consistent with this codebase's convention of using e2e/visual regression for layout components generically (no other file in `components/admin/layout/` has a unit test), so this isn't a new gap the PR introduces — but it does mean this fix ships effectively unverified until #777 lands, and if #777's scope or breakpoint assumptions drift, the wrap fix and its `test.fail()` flip could silently miss each other. Not blocking, but flag it so the two PRs are merged in the right order and someone actually re-runs the 1024 px check once both are in.

No CRITICAL or security issues. The `maxWidth: '100%'` approach and the `useFlexGap` addition on `PageHeader`'s own `Stack` are both correct fixes for the reported overflow.

## Review Summary

| Severity | Count | Status |
|----------|-------|--------|
| CRITICAL | 0     | pass   |
| HIGH     | 2     | warn   |
| MEDIUM   | 1     | info   |
| LOW      | 0     | pass   |

Verdict: WARNING — the overflow bug (#774) is genuinely fixed, but the wrapped action row's alignment (left instead of right) and `ReviewHubHead`'s missing `useFlexGap` (zero gap between wrapped button rows) should be addressed before this ships as the final word on #774, since both defects only become visible in exactly the 1024 px/open-sidebar scenario this PR targets.

bigpuritz:
Review addressed:

- **HIGH (wrapped action row left-aligned)** → the action slot carries `ml: { md: 'auto' }` now, so it stays right-aligned even as the sole item on a wrapped line, where `space-between` would park it at line start. Below `md` the stacked column keeps its left-aligned look.
- **HIGH (no vertical gap between wrapped rows in ReviewHubHead)** → its root Stack uses `useFlexGap` now, so the wrap this PR enables gets a real row gap instead of flush lines.
- **MEDIUM (verification deferred to #777)** → acknowledged; merge order stays fix-first by design, the net fails on an unexpected pass once it rebases.

PR #785, opened 2026-08-29
fix(ui): the title keeps its words, the adornment steps aside

### LINKED ISSUE #775: Review title collapses to one letter at 1440 px
Seen in the screenshot baselines of the responsive regression net (#725): on the review workspace at **1440 × 900** the review title "Smoke Review" renders as **"S…"** — the `h1` measures 45 px in a 286 px slot — while the head's middle segment (status pill, owner, progress meter, due date) and the 784 px action row keep their full width. At 1920 px the title gets 173 px, still an ellipsis for anything longer than two short words. 1440 is the most common laptop width reviewers work at, and the title is the one thing a head must show.

### Fix

Give the title a sane minimum (e.g. `minWidth: 200` or a `flex: 1 1 …` basis that beats the middle segment), let the middle segment wrap or hide its lower-priority items first (due date, then owner), and reserve the ellipsis for genuinely long titles. Not an overflow — the net cannot assert it with the overflow check — so verify by measurement or by a screenshot baseline update.

Part of #461.

bigpuritz:
## Summary

At 1440 × 900 the review workspace title rendered as "S…" — a 45 px h1 in a 286 px slot — while the middle segment (workflow milestones, badges) and the 784 px action row kept their full width (#775, seen in the responsive net's screenshot baselines #725).

Fix in `PageHeader`, on top of #774 (**this PR is stacked on `fix/issue-774-review-head-wrap` and retargets to `main` automatically once #784 merges**):

- The title row wraps (`flexWrap` + `useFlexGap`): when the slot gets tight, the `titleAdornment` steps onto its own line **below** the title instead of crushing the h1.
- The `noWrap` ellipsis now only ever hits a title that alone exceeds the slot — "reserve the ellipsis for genuinely long titles", as the issue asks. No magic `minWidth` constant needed: flex line-packing gives the title priority naturally.
- Together with #774's width-driven stacking the 1440 px head now resolves as: everything on one line when it fits → action row below when not → adornment below the title as the last step.

## Test plan

- [ ] Screenshot-baseline verification through PR #777 once both meet (this finding is not an overflow, so the net asserts it visually)

Closes #775

### DIFF
--- a/qnop-ui/src/components/admin/layout/PageHeader.tsx
+++ b/qnop-ui/src/components/admin/layout/PageHeader.tsx
@@ -74,7 +74,16 @@ export function PageHeader({
       <Stack direction="row" spacing={1.5} sx={{ minWidth: 0, alignItems: 'center' }}>
         {leading && <Box sx={{ flexShrink: 0 }}>{leading}</Box>}
         <Box sx={{ minWidth: 0 }}>
-          <Stack direction="row" spacing={1.5} sx={{ alignItems: 'center' }}>
+          {/* The title row wraps (issue #775): when the slot gets tight the
+              adornment steps below the title instead of crushing the h1 to a
+              few letters — the ellipsis is reserved for a title that alone
+              exceeds the slot. */}
+          <Stack
+            direction="row"
+            spacing={1.5}
+            useFlexGap
+            sx={{ alignItems: 'center', flexWrap: 'wrap' }}
+          >
             <Typography variant="h1" sx={{ fontSize: 28 }} noWrap={Boolean(titleAdornment)}>
               {title}
             </Typography>
--- a/qnop-ui/src/components/reviews/hub/ReviewPageHeader.tsx
+++ b/qnop-ui/src/components/reviews/hub/ReviewPageHeader.tsx
@@ -19,6 +19,7 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
+import Stack from '@mui/material/Stack';
 import type { AnnotationView, DocumentResponse } from '../../../api/generated';
 import { AnnotationStatus } from '../../../api/generated';
 import { PageHeader } from '../../admin/layout/PageHeader';
@@ -66,7 +67,10 @@ export function ReviewPageHeader({
       title={document.title}
       leading={<DocumentIcon size={34} contentType={document.contentType} />}
       titleAdornment={
-        <>
+        // One Stack, not a Fragment: the header's title row wraps (issue
+        // #775), and two loose flex items could split onto separate lines —
+        // milestones and badge travel as a single unit instead.
+        <Stack direction="row" spacing={1.5} sx={{ alignItems: 'center' }}>
           <WorkflowMilestones
             state={document.workflowState}
             total={annotations.length}
@@ -74,7 +78,7 @@ export function ReviewPageHeader({
             archivedAt={document.archivedAt ?? null}
           />
           {document.anonymous && <AnonymousBadge />}
-        </>
+        </Stack>
       }
       action={
         <ReviewHubHead


bigpuritz:
Reviewed the delta only (`fix/issue-774-review-head-wrap...fix/issue-775-review-title-min-width`), i.e. `qnop-ui/src/components/admin/layout/PageHeader.tsx` lines 74-91.

**Mechanics check (flexbox multi-line sizing):** the fix is sound. `Typography` already had `noWrap={Boolean(titleAdornment)}` before this diff, which sets `overflow: hidden` and therefore (per the CSS Flexbox spec's automatic-minimum-size rule) makes the item's `min-width: auto` resolve to `0` instead of its min-content. Combined with the new `flexWrap: 'wrap'` + `useFlexGap` on the inner `Stack`, this means: (1) line-breaking is decided on each item's max-content size, so `WorkflowMilestones`/badges (which are `white-space: nowrap`, no `overflow: hidden`) never shrink and instead get pushed to their own line when they don't fit; (2) once the title is alone on a line, the free-space-resolution step for that line only shrinks it down to the line's own width — so ellipsis genuinely only fires once the title text alone exceeds the full slot, exactly as the issue asked for ("reserve the ellipsis for genuinely long titles"). This also explains the original bug correctly: without wrapping, all shrink pressure landed on the only shrinkable item (the title), collapsing it to `"S…"`.

Compared to the issue's own suggested fix (`minWidth: 200` / a `flex-basis` that beats the middle segment / hiding lower-priority fields), this is a cleaner, dependency-free solution — no magic pixel constant, and it composes naturally with #774's outer-row wrap (title → actions → adornment, each stepping down only when needed).

**[MEDIUM] Compound `titleAdornment` may split across more than two lines — unverified**
`ReviewPageHeader.tsx` passes `titleAdornment={<>{"<WorkflowMilestones .../>"}{document.anonymous && <AnonymousBadge />}</>}`. Because it's a Fragment, `WorkflowMilestones` and `AnonymousBadge` become two independent flex items in the wrapping row, not one unit. At an intermediate width — where the title + `WorkflowMilestones` almost fit but the extra `AnonymousBadge` doesn't — this could produce a 3-line head (title / milestones / badge) rather than the clean "title, then adornment below" the PR description promises for the anonymous-review case. I can't confirm this visually without a browser render; worth including an anonymous review at ~1330-1420 px in the #777 screenshot-baseline pass to rule it out.

**[LOW] Acceptance criterion left for a follow-up PR**
The issue explicitly asks to "verify by measurement or by a screenshot baseline update" since the responsive net's overflow check can't assert this. The PR is candid about it (unchecked test-plan item, deferred to #777), which is the right call, but note that merging this PR will auto-close #775 (via "Closes #775") before that visual verification actually lands — worth keeping #777 linked/blocking so the fix doesn't get lost track of once the issue closes.

No CRITICAL or HIGH issues found. The other `titleAdornment` consumers (`HomePage`, `TeamDetailPage`, `MyTeamDetailPage`) carry small, single-element adornments and are effectively unaffected — they still won't wrap at realistic widths.

## Review Summary

| Severity | Count | Status |
|----------|-------|--------|
| CRITICAL | 0     | pass   |
| HIGH     | 0     | pass   |
| MEDIUM   | 1     | info   |
| LOW      | 1     | note   |

Verdict: APPROVE — sound, minimal, idiomatic flexbox fix for #775; one MEDIUM item worth a quick visual check in the promised #777 baseline pass before considering the responsive net fully green.

bigpuritz:
Review addressed:

- **MEDIUM (Fragment adornment = two loose wrap items)** → `ReviewPageHeader` now passes the milestones and the anonymous badge inside one `Stack`, so they wrap below the title as a single unit — the head resolves to two lines, never three.
- **LOW (visual acceptance deferred)** → acknowledged; asked PR #777 to add an anonymous review at ~1330–1420 px to its screenshot baselines so the verification cannot get lost once this issue auto-closes.

Issue #599, opened 2026-07-24
feat(review): composer-body extension point in the runtime UI extension model


bigpuritz:
## Motivation

The WYSIWYG editing mode is an enterprise feature (first consumer: qnophq/qnop-ee#2). The community owns the shared Markdown composer (#445) used by every annotation/comment surface. This issue adds the **composer-body extension point** to the runtime UI extension model (ADR-0039) so a registered module can contribute an alternative editing surface — without the community bundle shipping (or knowing about) any editor.

## Scope

- The composer's mode strip (`Write | Preview`) becomes extensible: a registered **composer-mode contribution** adds a mode tab and renders its own editing surface in place of the textarea while active.
- The contribution receives the controlled contract the textarea already honours: `value` (raw Markdown, stays the single source of truth), `onChange`, submit shortcut handling, `disabled`, fullscreen state, and the shared affordances (mention roster, attachment upload callback, emoji insertion) so a contributed mode can reach feature parity.
- Mode preference: the composer persists the last-used mode per user through the existing user-settings mechanism ONLY when an extension is registered; without extensions nothing changes visually or behaviourally.
- The Preview tab's visibility is owned by the active mode (a WYSIWYG-style mode may suppress it).

## Acceptance criteria

- [ ] Community behaviour and DOM are unchanged without a registered contribution (existing composer tests untouched).
- [ ] A test-only fake mode proves: tab appears, receives/updates the Markdown value, submit shortcut and disabled state flow through, roster/attachment callbacks are reachable.
- [ ] Storage format remains raw Markdown by contract — the extension point exposes no way to submit anything else.

Prerequisite for qnophq/qnop-ee#2. Follow-up to #445. ADR-0039 is the packaging frame.

🤖 Co-Author: Claude (Fable 5) via Claude Code

Issue #595, opened 2026-07-24
feat: enforce cross-namespace slug uniqueness (users and teams)


bigpuritz:
## Motivation

Team mentions (follow-up to qnophq/qnop#462) need `@<slug>` tokens to resolve unambiguously to exactly one principal. Today user slugs (#486) and team slugs (#470) are unique only within their own table — `@design` could be both a user and a team. Instead of a documented resolution-precedence rule, slugs become **one flat identity namespace**: a slug identifies exactly one principal, user or team.

## Current state

- `UserSlugService.allocate` / `TeamSlugService.allocate` are mirrored twins; each checks only its own table (`existsBySlugIgnoreCase`), backstopped by the per-table case-insensitive unique indexes `ux_qnop_user_slug_lower` / `ux_team_slug_lower`.
- The check-then-save race *within* one namespace is covered by those indexes; *across* namespaces nothing guards it.

## Scope

1. **Shared namespace check** used by both slug services: a candidate is free only when it exists in **neither** `qnop_user` **nor** `team` (case-insensitive). A small shared `SlugNamespace` helper keeps the two services from drifting apart.
2. **Close the cross-table race** with a transaction-scoped Postgres advisory lock (`pg_advisory_xact_lock(hashtext(lower(slug)))`) taken during allocation — Postgres-only is set by ADR-0020. A `slug_registry` table was considered and rejected (new maintenance surface for create/delete paths); see the ADR.
3. **Data migration** (additive Liquibase changeset, post-1.0 rule): resolve existing case-insensitive collisions between team and user slugs by re-allocating the **team** slug to its next free `base-n` candidate — user slugs are external identity (profile links, mention tokens in stored Markdown); team slugs are internal `/my-teams/…` links. Expected collision count: ~0. A re-slugged team breaks saved links (404) — accepted and documented in the ADR.
4. **ADR-0048** (cross-namespace slug uniqueness) lands in the same PR.

## Acceptance criteria

- [ ] Creating a user whose derived slug collides with an existing team slug allocates the next free `base-n` candidate instead — and vice versa, case-insensitively.
- [ ] Two concurrent allocations of the same slug across namespaces cannot both succeed (advisory-lock integration test).
- [ ] Migration re-slugs colliding teams; all changesets are additive.
- [ ] ADR-0048 added in the same PR.

Prerequisite for team mentions (qnophq/qnop-ee#1 — enterprise feature).

🤖 Co-Author: Claude (Fable 5) via Claude Code

Issue #462, opened 2026-07-11
@mentions in annotations and comments (feature + email notification)


bigpuritz:
## Why

Reviewers cannot direct a remark at a specific person. A `@username` mention is the natural way to pull someone into a discussion — and the notification that goes with it is the strongest "you are needed here" signal the system can send. Mentions do not exist in qnop today in any form: no parsing, no composer support, no rendering (verified against `main`).

Depends on #316, which builds the notification infrastructure (event bus, async mail dispatch, recipient resolution, per-user opt-out, `review.*` template namespace) that the mention mail plugs into.

### 1. Mention model (backend)

- Mention syntax `@username` in annotation opening comments and thread comments; resolved server-side on create against users **with access to the document** (owner, direct participants, team members). Unresolvable or out-of-roster mentions stay plain text.
- Persist resolved mentions per comment (join table), so rendering and notifications don't re-parse text.
- Note: OIDC users may have no `username` — decide fallback (mentionable via display name picker that inserts a canonical token, or not mentionable).

### 2. Composer autocomplete (frontend)

- Typing `@` in the Markdown composer opens a participant picker (roster of the document, not all users); selection inserts the canonical mention token.
- Rendering: mentions highlighted in `Markdown` output, linking to the user's profile page.

### 3. Anonymity (#413) — design decision required

In anonymous reviews, both the autocomplete roster and a rendered `@realname` would leak identities. Options to decide before implementation:
- Pseudonymous mentions: picker shows "Participant N" / "Reviewer team" tokens, server maps them to real users, rendering stays pseudonymous for everyone except self/owner (mirrors `ReviewIdentityResolver` semantics), or
- Mentions disabled for anonymous reviews (simplest, defensible v1).

### 4. Notification

- New event `ReviewUserMentioned` + template `review.mention` on the #316 pipeline (async, AFTER_COMMIT, opt-outs respected).
- Dedupe against `review.comment_added`: a mentioned thread participant gets **one** mail (the mention one), not two.
- Mention mail ranks above thread mail: subject names the actor and document, CTA deep-links to the comment (#412 permalinks).

## Out of scope

- Cross-document mentions, group mentions (`@team`), mention-driven auto-adding of participants (mentioning someone without access does nothing but render as text).
- In-app notification center (mail only, like #316).

## Acceptance criteria

- [ ] `@` autocomplete in composers offers exactly the document roster; canonical tokens inserted
- [ ] Mentions render highlighted and link to profiles; plain text when unresolved
[...]

PR #790, opened 2026-08-30
feat(review): pluggable mention resolution — seams for non-user mentions (#598)

### LINKED ISSUE #598: feat(review): pluggable mention resolution — seams for non-user mentions
## Motivation

Team mentions are an enterprise feature (first consumer: qnophq/qnop-ee#1). The community core must not contain team-mention logic — but it owns the mention machinery (#462): the server-side resolution (`CommentMentionService`), the composer's roster picker, the pill renderer (`MentionLink`) and the plain-text excerpt resolver (`useMentionNames`). This issue adds the seams so an enterprise module can extend WHO a mention token resolves to, without the community code knowing about teams.

Cross-namespace slug uniqueness (#595) guarantees a token resolves to at most one principal, so the seam never has to arbitrate ambiguity.

## Scope — server

- Extract the token→principal step of `CommentMentionService.resolveAndPersist` behind a **mention-resolver seam**: the community default resolves user slugs to the mentioned-user rows exactly as today (`@ConditionalOnMissingBean` / list-of-contributors pattern, ADR-0039); an enterprise contributor may resolve a token to a *set* of user ids (e.g. a team's members) with its own access rule.
- The persisted shape stays `comment_mention` user rows — the notification path (mention mail, opt-out, dedup) remains untouched by design.
- Decide whether the seam belongs in `qnop-spi` (published, versioned) or stays an internal Spring seam; ADR territory if `qnop-spi` grows a third contract.

## Scope — UI (runtime UI extension model, ADR-0039)

- **Roster extension point:** `useMentionRoster` accepts additional candidate groups contributed by a registered extension (with `kind`, avatar renderer, optional hint like "notifies N people").
- **Pill extension point:** `MentionLink`'s slug resolution falls through to registered resolvers before the raw-`@slug` fallback, so a non-user pill (avatar + name + link target) can be contributed.
- `useMentionNames` (excerpts, task titles, search) consults the same contributed name source.

## Acceptance criteria

- [ ] Community behaviour is byte-identical without an extension present (tests unchanged).
- [ ] A test-only fake contributor proves: extra roster group appears, its token resolves server-side to the contributed user ids, the pill renders through the contributed resolver.
- [ ] ADR if the seam lands in `qnop-spi`.

Prerequisite for qnophq/qnop-ee#1. Depends on #595. Follow-up to #462.

bigpuritz:
Closes #598. Prerequisite for qnophq/qnop-ee#1 (team mentions); depends on #595 (merged), follow-up to #462.

## Server — `MentionResolver` becomes qnop-spi's third published contract (ADR-0058)

- `io.qnop.spi.mention.MentionResolver` — `Set<UUID> resolve(MentionContext, slug)`: the user ids one `@slug` token addresses, empty when the resolver does not own the slug. `MentionContext` carries `documentId`, `ownerId`, `authorId`. Pure JDK, covered by the existing ArchUnit purity rule.
- **List-of-contributors wiring**, not `@ConditionalOnMissingBean`: the Community default `UserSlugMentionResolver` (profile slugs, case-insensitive — byte-identical to today) is always registered; an add-on contributes resolvers for other namespaces beside it. Slug uniqueness across namespaces (#595) keeps answers disjoint.
- Two invariants stay in the core, outside the seam: the **document-access rule** runs over every resolved id (a resolver names whom a token stands for, it can never widen who may be mentioned), and the persisted shape stays per-user `comment_mention` rows — mail/opt-out/dedup untouched. Anonymity (ADR-0038) is enforced before resolvers are called.
- `resolveAndPersist` now takes the comment's `authorId` (both call sites in `AnnotationService`).

## UI — mention contributor seam (in-process seed of ADR-0039)

- `qnop-ui/src/extensions/mentions.ts`: `MentionContributor` contract + registry (`useSyncExternalStore`), the germ of the future `qnop-ui-spi`; the ADR-0039 loader will feed it once it exists. #599/#600/#602 can reuse the pattern.
- `useMentionRoster` appends contributed candidates (kind, avatar, hint) after the user roster — never in anonymous reviews; the composer picker renders their hint and initials/own avatar (a contributed id is never fed to the user-avatar endpoint).
- `MentionLink` consults contributors before the user-profile fetch (query disabled when a contributor owns the slug) and renders the contributed pill — name, avatar, link target, no player hover card.
- `useMentionNames` resolves contributed slugs in plain-text excerpts, after roster names.

## Acceptance criteria

- [x] Community behaviour byte-identical without an extension (existing tests unchanged and green; explicit no-contributor tests added)
- [x] Test-only fake contributor proves: extra roster candidates appear, its token resolves server-side to the contributed user ids (union, dedup, access-filtered), the pill renders through the contributed resolver
- [x] ADR-0058 for the third qnop-spi contract; ARCHITECTURE.md/CLAUDE.md updated

## Test plan

- [x] `CommentMentionServiceTest` (incl. fake team resolver) + `ArchitectureRulesTest` green
- [x] `:qnop-spi:build` green
- [x] Frontend: 1388 tests, `tsc`, `eslint`, `prettier` clean
- [ ] CI green (incl. Docker ITs — `MentionNotificationIT` unchanged by design)


bigpuritz (commented):
## Independent agent review
Scope: full PR #790 diff against `main` (backend `MentionResolver` SPI + `CommentMentionService` rewiring + frontend mention-contributor registry), reviewed with no prior context; claims verified against the code.

**Verdict: Approve with comments.** The core design is sound and the security-relevant claims hold up under verification: the document-access rule (`hasAccess`) genuinely runs over every resolver-returned id inside the core, so a resolver cannot widen who is mentioned or notified (proven by `accessRuleAppliesToContributedIdsToo`); the ADR-0038 anonymity check runs before any resolver is called; `UserSlugMentionResolver` reproduces the old `findBySlugIgnoreCase` lookup verbatim, and with no contributor registered every frontend path is identical to before. The new `io.qnop.spi.mention` package is automatically covered by the existing ArchUnit purity rule (`.definedBy("io.qnop.spi..")`). No CRITICAL findings. One HIGH (a false statement in the published contract's javadoc), two MEDIUM, three LOW.

### HIGH

- **H1 — Published contract javadoc is factually wrong about case normalization.** `qnop-spi/src/main/java/io/qnop/spi/mention/MentionResolver.java:47-49` says: "`slug` is passed as written (case not normalized); resolvers match it case-insensitively". But the core lowercases every token before resolvers ever see it — `MentionParser.extractSlugs` does `matcher.group(1).toLowerCase(Locale.ROOT)`, and `resolveAndPersist` iterates those lowercased slugs. So the slug is *never* passed as written. The two sentences contradict each other, and in a published, semver'd artifact the javadoc *is* the contract: an implementer who takes "as written" at face value and exact-matches against display-cased slugs will silently never resolve anything (the test `TeamResolver` masks this by using `equalsIgnoreCase`). Fix the javadoc before this ships in an artifact — either "the slug arrives lowercased (`Locale.ROOT`)" (matching reality, and arguably a useful guarantee to promise) or actually pass it as written. Note ADR-0058 and the #792 EXTENSION-POINTS entry repeat the case-insensitive phrasing and should stay in sync.

### MEDIUM

- **M1 — Anonymity gate is asymmetric across the three frontend consumers.** `useMentionRoster.ts` correctly returns `[]` for anonymous reviews before consulting contributors (and has a test for it). But `MentionLink.tsx:60` and `useMentionNames.ts` consult contributors unconditionally, with no anonymity check. In an anonymous review the server leaves mentions unresolved (ADR-0038) and user slugs stay raw in the UI (the anonymized roster exposes no slugs to the `names` map), yet a contributed slug typed as plain text — `@platform-team` — would render as a fully resolved pill/name client-side. That contradicts both the server's "tokens stay plain text for everyone" behavior and the PR's own "never in anonymous reviews" framing, and it advertises which teams exist/are pingable inside a review that is deliberately anonymous. Suggest threading the review's `anonymous` flag into `MentionLink`/`useMentionNames` (or having them skip `resolveContributedMention` when the hosting review is anonymous), plus a test.

- **M2 — No guard around resolver calls in the core loop.** `CommentMentionService.resolveAndPersist` (qnop-core/.../CommentMentionService.java, new resolver loop): a contributed resolver that throws — despite the javadoc's "must not throw" — propagates out of the `@Transactional` method and aborts the whole comment/annotation creation (`AnnotationService.create` / `addComment` both call it inline). A resolver returning `null` (contract violation, but nothing enforces it) NPEs in the enhanced-for the same way. Under the trusted-extension model (ADR-0049) failing loud is defensible, but "one buggy mention add-on makes all commenting throw 500s" is a harsh blast radius for a purely additive feature. At minimum this failure mode deserves a deliberate, documented decision; a per-resolver try/catch that logs and skips (or an `Objects.requireNonNullElse(..., Set.of())`) would confine the damage to the contributed namespace.

### LOW

- **L1 — Case-sensitive dedup in the roster.** `useMentionRoster.ts:92` dedups contributed candidates with `candidate.slug === principal.slug` while every other slug comparison in the codebase is case-insensitive. A contributor supplying `Platform-Team` next to an identically-slugged entry differing only in case produces a duplicate picker row. Compare lowercased.

- **L2 — Frontend contract omits case semantics.** `mentions.ts` — `MentionContributor.resolve(slug)` documents nothing about case handling; both test fakes quietly implement `slug.toLowerCase()` themselves, and `useMentionNames`/`MentionLink` pass the slug through unnormalized. Document (or normalize before calling) so an extension author gets the same guarantee the backend SPI states.

- **L3 — `candidate.kind` truthiness as the user/non-user discriminator.** `MarkdownComposer.tsx:601` routes avatar sourcing on `candidate.kind ? contributedAvatar : avatarSrc(candidate.id)` — a contributor setting `kind: ''` would feed a non-user principal id to the user-avatar endpoint, the exact thing the comment forbids. `candidate.kind !== undefined` (or a dedicated `isContributed` flag) is safer. Also: the picker's hint/avatar rendering added here has no direct test — `mentionExtensions.test.tsx` covers roster/pill/names but not the composer menu item.

### Test coverage
Good overall: both acceptance criteria are genuinely tested (fake contributor end-to-end on both sides, explicit no-contributor "byte-identical" tests, access-rule and anonymity tests, registry subscribe/notify semantics, and the `useUserProfile(slug, false)` disabled-query assertion the PR body claims). Gaps worth closing: M1's anonymous-review pill/excerpt behavior, and L3's picker rendering.

bigpuritz:
The independent-review findings on this (merged) PR are addressed in the follow-up https://github.com/qnophq/qnop/pull/796 — see its table for the finding-by-finding mapping, including the rebuttal of the anonymity-gate MEDIUM (rendering resolves public identities exactly as user pills already do; ADR-0038 is enforced at creation/notification, not at render). 🤖 Co-Author: Claude (Fable 5) via Claude Code

Issue #689, opened 2026-08-02
[epic] Extension seams for the enterprise integration wave


bigpuritz:
## What this covers

Three enterprise features have been scoped in the private repository — guest links (qnop-ee#17), webhooks (qnop-ee#18) and machine API access (qnop-ee#19). All three ship as extensions, and none of them can attach today: an in-process extension reaches the core only through published seams (ADR-0049), and the seams they need do not exist.

This epic tracks the core-side prerequisites. **The features themselves stay in the enterprise repository**; what lands here is the vocabulary the core has to grow so an extension can supply behaviour without reaching into internals.

- **#684** — a review participant without an account (for qnop-ee#17)
- **#685** — a published event contract extensions can subscribe to (for qnop-ee#18)
- **#686** — letting an extension contribute an authentication mechanism (for qnop-ee#19)

## The overlap worth seeing before any of them starts

**#684 and #686 are two faces of the same gap.** The core knows exactly one kind of actor: a `User` row. A guest is an actor without a user account. A service account is an actor without a human behind it. Both need the same thing from the core — *whoever is acting is not necessarily a user record* — and both currently break on the same places: the security filter chain has no non-account principal, and `AuditEvent.actorId` assumes one.

Solved separately, this produces two principal concepts that mean almost the same thing, and every later feature has to ask which one it is dealing with. Solved once, it is a single abstraction with two suppliers — which is also the only way either seam gets the second consumer that ADR-0049 says extension points need in order to be cut correctly.

**Recommendation: settle the principal question once, across #684 and #686, before either is designed in isolation.** #685 is genuinely independent and can proceed on its own.

## Suggested order

1. **#685 (events)** — the most self-contained, and the one whose scope is already clear. Good place to establish the pattern the other two follow: what a published seam looks like, how it is versioned, how it is tested without an enterprise build.
2. **#686 (authentication)** — starts with a question that may shrink it dramatically: the core already runs two `SecurityFilterChain` beans, so an extension can probably add a third for its own paths with no core change. Answer that first; it decides whether there is a seam here at all beyond the shared principal work.
3. **#684 (account-less participant)** — the largest, and the one that most needs the principal decision from step 2 to exist first.

## Constraints that apply to all three

- **The core learns no commerce.** No edition checks, no entitlements, no awareness that a consumer is paid for (ADR-0003). Each seam is a capability, not a product boundary.
[...]

Issue #422, opened 2026-07-05
Anonymous review leaks the reviewer roster to non-owners; add an anonymity marker


bigpuritz:
Follow-up to #413 (per-review anonymity, PRs #419/#420). Anonymity currently hides *authorship* on annotations/comments (foreign authors render as `P1`, `P2`, …), but the **reviewer roster is still shown in full to everyone** — which defeats the point of an anonymous review. This revises the "the participant roster stays visible" scoping note in ADR-0038.

## Bug — the participant roster leaks in an anonymous review

In an anonymous review, **only the owner** should be able to see who the participants are. Every other participant should see the same anonymised labels (`P1`, `P2`, …) everywhere — consistent with how authorship already renders.

Today, in an anonymous review, the full reviewer list is exposed to all participants in:

- **The reviews overview** — the review cards/table show every participant's avatar (`DocumentSummary.participants`).
- **The review details view** — the header participant avatars and the "Participants" dialog (`GET /documents/{id}/participants`, and the team-member unfold `GET /principals/teams/{id}/members`) show real names.

Because this is a privacy guarantee, it must be **enforced server-side** (like the rest of #413) — the roster APIs must not ship real identities (names, or resolvable principal ids) to a non-owner caller when the review is anonymous. Hiding it only in the client is not enough.

### Design points for planning

- **How to represent the roster to a non-owner** in an anonymous review: anonymised labels (`P1`, `P2`, …) vs. a bare count ("4 reviewers") vs. showing only the viewer themselves + the owner. The user's intent is anonymised labels consistent with the annotation pseudonyms.
- **Consistency of the pseudonyms.** The annotation/comment pseudonym ordinal (`ReviewIdentityResolver`) is currently assigned over *non-owner authors*, but the roster includes reviewers who have not authored anything, and **teams**. Decide how a roster entry maps to a stable label so `P1` in the roster is the same person as `P1` on their notes (extend the ordinal to cover all participants? teams too?).
- **The owner is exempt** (as elsewhere) and sees the real roster; the owner's own identity stays visible in the header.

## Enhancement — visually mark anonymous reviews

An anonymous review should carry a clear visual indicator so every participant immediately knows they are in one (and that names are hidden). Suggested surfaces:

- A badge/chip in the **review header** (next to the workflow-state chip), e.g. an "Anonymous" pill with an eye-off icon.
- A matching marker on the **review card** in the overview.

## Acceptance

[...]

Issue #413, opened 2026-07-05
Per-review privacy: anonymous reviews and thread participation policy


bigpuritz:
Today, foreign annotations and comments render under the anonymous label **"Participant"** — you cannot see who actually wrote them. That is currently an accident of the UI (author names are simply not resolved everywhere) rather than a decision. It should become a **per-review configuration, chosen when the review is created**, and enforced by the server — not just cosmetically in the client.

## Setting 1 — review anonymity

- **Anonymous review**: no personal names anywhere — every foreign author renders as an anonymised "Participant" (the viewer's own contributions still read "You"). The API must not ship author names or resolvable author identities for display in this mode; anonymity that only lives in the frontend is none.
- **Non-anonymous review**: real display names everywhere — annotation heads, comment bubbles, task-board cards, author filter, hover preview. Server ships `authorDisplayName` on the annotation/comment views (resolution server-side, since an author may participate via a team and never appear in the participant rows).

Open design points for planning: whether anonymous mode uses one uniform label or stable pseudonyms ("Participant 3") so threads stay followable; whether the owner is exempt from anonymisation.

## Setting 2 — thread participation policy

Who, besides the annotation's **author** and the **owner**, can take part in an annotation's thread? Three levels:

| Level | Foreign threads are… |
|-------|----------------------|
| `PRIVATE` | not visible at all (annotations/comments filtered out server-side) |
| `READ_ONLY` | visible, but only author and owner may comment |
| `OPEN` | visible and writable for every participant (today's behaviour, the default) |

Enforced at the API (annotation/comment listing and `addComment` guard), mirrored in the UI (panel, focus, tasks, hover preview).

## Consequences to handle

- **FINALIZED gate**: the zero-OPEN invariant counts *all* annotations — under `PRIVATE`, reviewers may be blocked by annotations they cannot see; surface that comprehensibly (e.g. "3 open annotations, 2 not visible to you").
- **Tasks board / progress bars**: counts follow the caller's visibility under `PRIVATE`.
- **Permalinks (#412)**: a link into an invisible thread answers the usual 404-style degradation.
- **Likes (#410)** and the author facet of the filter bar respect anonymity (no author facet in anonymous mode).
- Both settings live on the document, are set in the "New review" wizard, and are immutable after creation for now (changing them mid-review rewrites history semantics — own topic if ever needed). Pre-prod: additive columns in the existing changeset.

## Acceptance

[...]

PR #797, opened 2026-08-31
feat(core): published review-event stream — curated catalogue in qnop-spi (#685)

### LINKED ISSUE #685: seam: publish an event contract extensions can subscribe to
## Why

Webhooks are being built as an enterprise extension (qnop-ee#18): forward qnop events to an external endpoint. An extension attaches only through published seams (ADR-0049), and today an extension has no supported way to learn that anything happened. **This issue is the core-side prerequisite** — the event contract, not the delivery machinery. Retry, signing, SSRF checks and the subscription UI stay in the extension.

## The obvious answer looks closer than it is

`ReviewNotificationSink` is already an interface with two implementations, so "publish it into `qnop-spi`" looks like a five-minute job. It is not, and the reason is worth writing down before someone tries:

```java
boolean accepts(ReviewNotificationIntent intent);
void deliver(ReviewNotificationIntent intent);
```

The sink is **per recipient**. `ReviewNotificationService` decides who should hear about an event and offers each recipient's ranked candidates to the sinks in turn; a sink answers only for its own prerequisites and opt-outs. That design is right for channels that deliver *to a person* — in-app, mail.

A webhook is not that. It fires **once per event**, not once per interested human, and it has no recipient in the sink's sense. Reusing the interface would mean a sink that answers `accepts` arbitrarily and deduplicates deliveries by hand — a channel pretending to be a recipient. The mismatch would show up as duplicate webhooks the first time two participants are interested in one annotation.

So the seam webhooks need is **event-shaped, not recipient-shaped**, and it is a different one from `ReviewNotificationSink`. Whether that new seam should also absorb the existing sinks, or sit beside them, is an open question.

## What is available to expose

- **`AuditEvent`** is the broadest existing stream: a dotted `eventType`, an `actorId`, a `scope` separating `DOCUMENT` from `SYSTEM` (ADR-0043), a `detail` payload and a timestamp. Every audited action is already an event.
- **`NotificationType`** enumerates the review-side events — annotation created, comment added, annotation decided, version uploaded, workflow changed, participant added, mention, review deleted.

**What is the published event vocabulary?**

- *Expose the audit trail.* Nothing new to build; `eventType` strings become the contract. But those strings were written for operators reading a compliance log, not as an API — and publishing them freezes them. Renaming `scheduler.job.dry_run` would then be a breaking change.
- *A curated event catalogue.* Its own stable names, versioned deliberately, mapped from internal events. More work, ages far better, and lets internal refactoring stay internal.

The second is very likely right, but it needs someone to decide which events are in it — and that list should be short at the start, because removing one later is a breaking change and adding one is not.

## Constraints on any answer

- **No payload content in the contract.** A published event says *what happened to which object*, with identifiers. Annotation bodies and document titles are the customer's confidential material; whether a consumer may see them is a permission question answered by the API, not by an event bus. See qnop-ee#18 and qnop-ee#19.
- **Subscription must not be able to slow or break the action that produced the event.** Whatever the seam is, an extension's handler runs outside the originating transaction — that boundary belongs in the core, not in the extension's good behaviour.
- The core must remain unaware of webhooks, HTTP or delivery. It publishes events; who listens is not its concern.

## Definition of done

- A decision, recorded as an ADR, on the event vocabulary and on the relationship between the new seam and `ReviewNotificationSink`.
- The seam and its event types published in `qnop-spi`.
- A test-only listener in the core proving events arrive without an enterprise build.
- Demonstrated isolation: a listener that throws or blocks does not affect the originating request.

bigpuritz:
Closes #685 — first of the three seams of epic #689. Prerequisite for qnop-ee#18 (webhooks); the delivery machinery (retry, signing, SSRF, subscription UI) stays enterprise.

## Decision (ADR-0059)

**A curated catalogue, not an exposed internal stream.** `io.qnop.spi.event` is qnop-spi's fourth contract:

- `PublishedEventListener` — `on(PublishedEvent)`; every registered listener hears every event.
- `PublishedEvent` — `type`, `occurredAt`, `documentId`, `actorId`, identifier-only `attributes`.
- `PublishedEventTypes` — eight stable names (`review.annotation.created` … `review.deleted`) mapped from the sealed `ReviewEvent` hierarchy. Adding a name is minor; removing one is breaking — nothing enters without a consumer. The mapper is **exhaustive over the sealed hierarchy**, so a new internal event is a compile error until someone decides its published fate.

**Why not the obvious answers** (per the issue's own analysis): `ReviewNotificationSink` is recipient-shaped — a webhook is event-shaped; reusing it means hand-rolled dedup and duplicate deliveries. The audit trail's `eventType` strings are operator-facing and would freeze. Both rejected in the ADR; the sinks stay beside the new seam, not absorbed.

**Isolation demonstrated, not documented:** `AFTER_COMMIT` + dedicated bounded async executor (full queue ⇒ **drop with warning**, deliberately not CallerRuns — an extension must never slow the committing request) + per-listener wrapping. A throwing listener costs itself; siblings and the caller are unaffected (unit-tested).

**No payload content:** identifiers only; `ReviewDeleted.title` is explicitly dropped in the mapping (tested).

## Definition-of-done mapping

- [x] ADR-0059: vocabulary decision + relationship to `ReviewNotificationSink`
- [x] Seam + event types published in `qnop-spi` (ArchUnit purity rule covers the new package)
- [x] Test-only listeners in the core prove events arrive without an enterprise build
- [x] Isolation: throwing listener affects neither siblings nor the originating action (test); executor policy keeps the request thread out of it
- [x] `docs/EXTENSION-POINTS.md` §4 + contract-count updates (same-PR rule)

## Test plan
- [x] `PublishedEventDispatcherTest` (catalogue mapping incl. no-title, isolation, order, no-listener zero cost)
- [x] `:qnop-spi:build`, `ArchitectureRulesTest`, Spotless
- [ ] CI green

Part of epic #689.


bigpuritz (commented):
## Independent agent review
Scope: full PR #797 diff against `main` — the `io.qnop.spi.event` contract, `PublishedEventDispatcher`/`PublishedEventMapper`, the new executor, ADR-0059 and the EXTENSION-POINTS.md §4 entry — verified against the sealed `ReviewEvent` hierarchy, the existing async/notification plumbing, and the renumbered catalogue.

**Verdict: Approve with comments.** The design claims verify: the mapper's `switch` is genuinely exhaustive over all eight sealed `ReviewEvent` members with no `default` (a ninth member is a compile error); `ReviewDeleted.title` is the only content-bearing field in the hierarchy and it is dropped and tested; the per-listener try/catch is correct and tested; `AFTER_COMMIT` + `@Async` on a named executor is the same proven pattern as `ReviewNotificationListener`; the drop-not-CallerRuns rejection handler does not throw, so a full queue can never block or fail the committing thread — and the ADR's contrast with the sibling executor is accurate (I verified `reviewNotificationExecutor` really uses `CallerRunsPolicy`). The §-renumbering is complete: old §4–§6 shifted correctly, the `useMentionNames.ts` §5→§6 pointer was caught, in-document references (§3 contrast, "§4, ADR-0059" in the planned table) are all correct, and no stale references remain elsewhere. ArchUnit purity covers the new package via `io.qnop.spi..`. One HIGH, one MEDIUM, four LOW.

### HIGH

- **H1 — ADR-0038 anonymity is never considered, and the seam's own conventions table says it must be.** Every `PublishedEvent` carries the raw `actorId` — including events raised inside anonymous reviews, where the internal consumers of the very same stream go to careful lengths to hide exactly this: `ReviewNotificationService.actorNameFor` renders the actor pseudonymously per recipient (issue #413), and `InAppNotificationSink` documents that "nothing here can leak an identity". The published stream hands the real actor to extensions with no marker that the review is anonymous and no stated obligation — and the first consumer is a webhook forwarder (qnop-ee#18) whose whole purpose is to ship these events *out of the process*, where ADR-0038's promise to participants would be silently pierced. Worse, this contradicts the project's own written invariant: EXTENSION-POINTS.md's conventions table ("Invariants stay core-side") lists "anonymity (ADR-0038)… enforced by the core *around* the seam", and the mention seam (§3) honors that by never calling resolvers in anonymous reviews — §4 is now the one seam where the claim is false and the divergence is unmentioned. Trusted-code status mitigates the mechanics (an in-process extension could read the DB anyway) but not the contract: the deliberate decision is missing. Options, in descending preference: (a) suppress or blank `actorId` for events from anonymous reviews (needs the anonymity flag at publish time — the internal events carry ids only, so this is a mapper/dispatcher lookup or a flag on `ReviewEvent`); (b) add an `anonymousReview` attribute/flag so a consumer *can* comply; (c) at bare minimum, document in `PublishedEventListener`/`PublishedEvent` javadoc, EXTENSION-POINTS §4 and ADR-0059 that events from anonymous reviews carry the real actor id and the consumer owns ADR-0038 compliance before any egress — and amend the conventions-table row so it stops overpromising. As it stands, ADR-0059 models itself on the notification path yet omits the one policy that path enforces most carefully.

### MEDIUM

- **M1 — Ordering and concurrency semantics of a published contract are undocumented.** `publishedEventExecutor` runs 1 core / 2 max threads: `ThreadPoolExecutor` only spawns the second thread once the 500-slot queue is full, at which point a listener will be invoked *concurrently* from two threads and events will be delivered *out of order* — including two events of the same document. The javadoc says "safe to call from any thread" (covers re-entrancy) but nothing about ordering, and the test `deliversToEveryListenerInRegistrationOrder` can be misread as an ordering promise (it is listener order within one event, not event order). A webhook consumer that assumes per-document ordering (a natural assumption for `review.workflow.changed`) will be wrong precisely under load, the hardest case to reproduce. Either document "no ordering guarantee, events may arrive concurrently" in `PublishedEventListener` (and ADR-0059), or pin the pool to one thread and document best-effort FIFO — but decide it in the published text, not by executor accident.

### LOW

- **L1 — `qnop.events.sync-dispatch` is shipped dead.** The sync-executor bean's javadoc says "for tests that assert on delivered events", but no test sets it — `AbstractIntegrationTest` flips only `qnop.notifications.sync-dispatch`. Until a test uses it, it is untested config; and note that in the current ITs each review action now enqueues a real async no-op task (the `listeners.isEmpty()` early-return runs on the dispatcher thread, so "the mapping is skipped entirely" is accurate but "zero work" is not quite). Cheapest fix: set the property in `AbstractIntegrationTest` alongside its sibling, which also keeps stray dispatcher tasks from racing the per-test TRUNCATE.
- **L2 — `occurredAt` is stamped when the async task runs, not when the action committed.** With a backlog (up to 500 queued), the timestamp drifts arbitrarily far from the commit it describes. Acceptable for best-effort, but the `@param` text ("when the event was published (after the originating commit)") reads more precise than it is — one clause about queue-delay drift would keep a webhook consumer from treating it as a commit timestamp.
- **L3 — The drop warning identifies nothing.** "published-event queue full — dropping one event dispatch" says neither type nor document, and sustained drops produce only a warn-per-event stream. Fine for the seam's contract; an operator-facing counter (or at least occasional queue-state context) would make a misbehaving listener diagnosable. Optional.
- **L4 — `ParticipantAdded` user branch untested.** The mapping test covers the `teamId` variant only; the `userId` branch of the null-guarded pair is one assertion away from covered.

### What is genuinely good
The curated-catalogue decision (stable names, string `type` for forward-compat, "consumers must ignore unknown types" written into the contract, compile-error-on-new-internal-event via sealed exhaustiveness) is exactly how a semver'd event vocabulary should be built — growth is minor, nothing enters by accident. `PublishedEvent`'s compact constructor with `Map.copyOf` keeps the record genuinely immutable, and the rejected-execution lambda not throwing is the small detail this design lives or dies on — it was gotten right.

bigpuritz:
Review addressed in the follow-up commit:

- **HIGH (anonymity):** the dispatcher now nulls `actorId` for anonymous reviews before any listener hears the event — flag from the document, carried by `ReviewDeleted` itself, unloadable document ⇒ anonymous (erring toward silence). Contract javadoc, ADR-0059 and EXTENSION-POINTS §4 all state it; tested for all three paths.
- **MEDIUM (ordering):** documented as a contract property — no ordering guarantee, consumers reconstruct order from their own data; ADR records why single-thread pinning was rejected.
- **LOW (dead property):** `AbstractIntegrationTest` now sets `qnop.events.sync-dispatch=true` like its notifications sibling.
- **LOW (occurredAt):** javadoc now says dispatch time and names the backlog drift.
- **LOW (drop warning):** kept as is — the rejection handler sees only the Runnable, so type/document aren't available there; the warning documents that.
- **LOW (userId branch):** test added.

bigpuritz (commented):
Scope: re-review of PR #797 after commit 56c8db74, which addresses my six findings; each resolution re-verified against the code, with particular scrutiny on the anonymity gate.

**Verdict: five of six findings resolved and the gate itself is correctly built — but auditing every field that crosses the seam against ADR-0038 shows the fix closed the `actorId` path while leaving one sibling path open. One remaining HIGH.**

### Verified resolved

- **M1 (ordering): resolved.** "No ordering guarantee — two events, even of one action, can arrive concurrently or out of order" is now in the `PublishedEventListener` contract, and ADR-0059's new section states it with an honest rationale for not pinning the pool. Exactly what was asked: decided in the published text, not by executor accident.
- **L1 (sync-dispatch): resolved.** `AbstractIntegrationTest` now flips `qnop.events.sync-dispatch` beside its notifications sibling.
- **L2 (occurredAt): resolved.** The `@param` now says dispatch-time and names the queueing-delay drift.
- **L4 (ParticipantAdded userId branch): resolved.** `mapsTheUserVariantOfParticipantAdded` covers it.
- **L3 (drop warning): kept as-is with reasoning — fine.** It was optional; accepted.

### The anonymity gate — what is right

The mechanism is sound in all the places I probed for holes. ADR-0038 makes `anonymous` **immutable after creation**, so the post-commit lookup cannot race a toggle; the only async race is deletion, where `findById` → empty → counts-as-anonymous errs to silence, which is the correct direction. `ReviewDeleted` carrying its own flag is right (read from the aggregate before the row dies, inside the deleting transaction). The no-listener early return still precedes the repository read, and `verifyNoInteractions` now proves the zero-cost claim. Nulling even the owner's `actorId` over-suppresses slightly relative to ADR-0038's owner exemption (the owner is the publicly named coordinator) — a defensible simplification in the safe direction, no objection. Contract (`@param actorId … or null in an anonymous review`), ADR section and EXTENSION-POINTS §4 all match the code.

### HIGH — remaining leak: `review.participant.added` publishes real principal ids in anonymous reviews

The fix nulls `actorId`, but the **attributes** of `PARTICIPANT_ADDED` still carry `userId`/`teamId` raw, anonymous or not (`PublishedEventMapper` rebuilds the event with `published.attributes()` unchanged). ADR-0038's amendment for #422 hides exactly these: "in an anonymous review only the owner (and an admin) see who the reviewers are … with principal ids replaced by synthetic tokens (team tokens derive from position, **not the real team id**)" — enforced on the participants endpoint and the overview precisely because a readable roster "defeats the anonymity". A webhook consumer subscribed to an anonymous review therefore receives the full reviewer roster, principal by principal, as it is assembled — the one list the server everywhere else refuses to reveal, and off-process at that. Fix options: for anonymous reviews, strip the `userId`/`teamId` attributes (leaving "a participant was added"), or suppress `PARTICIPANT_ADDED` entirely there; either wants a test beside `anonymousReviewsNeverPublishTheActor`. Worth a one-line sweep note in the ADR too: in an anonymous review, *no identity — actor or subject —* crosses the seam, so a future event type with a principal-bearing attribute (e.g. a participant-removed event) inherits the rule instead of re-fighting it. (`ownerId` in `review.deleted` is fine — the owner is structurally public under ADR-0038.)

bigpuritz:
Second-pass finding fixed: in anonymous reviews the mapper now strips identity-bearing attributes (`userId`/`teamId`) along with the actor. The general rule — no identity, actor or subject, crosses the seam; `ownerId` exempt as structurally public — is recorded in ADR-0059 and EXTENSION-POINTS §4, with a dedicated test. 🤖 Co-Author: Claude (Fable 5) via Claude Code

bigpuritz (commented):
Scope: verification of commit 16ee9644 against the remaining HIGH from my second-pass review of PR #797.

**Verdict: resolved — Approve. All findings from both review passes are now addressed.**

- **The strip is complete.** `IDENTITY_ATTRIBUTES = {"userId", "teamId"}` covers exactly the principal-bearing keys in the current catalogue; I re-checked every other attribute (`annotationId`, `commentId`, `versionNumber`, `fromState`/`toState`/`manual`, `reopened`, `ownerId`) and none carries a hidden identity. The `ownerId` exemption is correct under ADR-0038 — the owner is the publicly named coordinator even in an anonymous review.
- **The mechanics are right.** The mutable working copy is re-frozen by `PublishedEvent`'s compact-constructor `Map.copyOf`, so the published event stays immutable; the strip runs in the same anonymity branch as the actor null, so the two can never diverge.
- **The rule is now general and recorded.** ADR-0059 states "no identity — actor or subject — crosses the seam in an anonymous review" with the ownerId exemption and the instruction that a future principal-bearing event type adds its key to the stripped set; EXTENSION-POINTS §4 says the same. Key-based stripping does mean a future attribute under a new name must remember to enroll — but with the rule written into the ADR, the contract docs, and the `IDENTITY_ATTRIBUTES` javadoc, that is a documented maintenance point, not a gap.
- **Tested.** `anonymousReviewsStripSubjectIdentitiesToo` asserts both the null actor and the empty attribute map for `participant.added` under anonymity, sitting beside the actor-null, deletion-flag and unloadable-document tests.

One immaterial nit, no action needed: the ADR sentence says "the dispatcher nulls `actorId` and strips…" — the stripping lives in the mapper; the dispatcher supplies the flag. The behavior described is accurate.

Nothing further from my side — a well-run three-pass exchange.

Issue #684, opened 2026-08-02
seam: a review participant without an account


bigpuritz:
## Why

Guest links — inviting an external reviewer without an account — are being built as an enterprise extension (qnop-ee#17). An in-process extension attaches only through published seams (ADR-0049), and there is none for a participant who has no `User`. **This issue is the core-side prerequisite, not the feature.** The feature stays in the enterprise repository; what belongs here is the concept the core has to understand before an extension can supply one.

Related: qnop#675 (closed, moved to qnop-ee#17) collected the security requirements such a feature has to meet.

## What assumes an account today

- **`ReviewParticipant`** is keyed to a `User`. Half the review paths resolve identity through it.
- **The security filter chain** (`SecurityConfiguration`) has no notion of a principal that is not a user. Filters are wired explicitly in one `@Bean` method; there is no extension point.
- **`ReviewIdentityResolver`** assigns "Participant N" pseudonyms by ascending **user id** for anonymous reviews (ADR-0038). No user id, no ordinal.
- **`thread_participation`** (`PRIVATE | READ_ONLY | OPEN`) is the vocabulary that already expresses what a participant may do — a guest's permission level should reuse it rather than introduce a second scheme.
- **`ReviewVisit`** records who looked at a review; guest visits either land there or the owner's picture of "who saw this" quietly acquires a hole.

## The question to settle first

**How far does the core have to go?** The extremes are both wrong:

- *Too little:* a bare "non-user principal" hook, leaving the extension to reimplement participation, identity resolution and visit tracking against internals it should not touch. The seam would leak.
- *Too much:* the core grows a full guest model — tables, lifecycle, invitation flow — and the enterprise feature becomes a thin UI over it. Then the AGPL core carries the substance of a paid feature, which defeats the point of putting it in the enterprise repository.

The useful middle is probably: the core learns that a **participant may be identified by something other than a user id**, and the extension owns *what* that something is, how it is issued and how it expires. Concretely that likely means a nullable user reference plus a stable opaque participant identity, an identity-resolution path that does not assume a user row, and a documented way to contribute an authentication mechanism. But that is a hypothesis, not a decision — it is what this issue exists to settle.

## Constraints on any answer

- The core must not learn anything about links, invitations, expiry or commercial gating. Those are the extension's business (ADR-0003).
- Anonymous reviews must stay anonymous with account-less participants present (ADR-0038), and nothing an extension supplies as a display name may break that.
[...]

Issue #21, opened 2026-06-14
OIDC/OAuth: DB-configured providers, login, auto-provisioning, admin


bigpuritz:
## Goal
DB-configured OIDC/OAuth login + auto-provisioning + admin provider management.

## Scope
- `DbClientRegistrationRepository` (rebuild on provider change), provider registry / registration-ids (= provider UUID), type-specific principal adapters (OIDC/Google/GitHub/Facebook/generic), `GitHubEmailFetcher`.
- `OidcLoginSuccessHandler` (issues refresh cookie + stores upstream id_token), `OidcIdentityService` auto-provision/link strictly by (provider, subject); no cross-provider linking.
- Resource-server per-provider decoders + validators (timestamp/issuer/audience); RP-initiated logout (`end_session_endpoint`).
- `OidcSsrfPolicy` (DNS-free host classifier; blocks private/loopback/metadata; opt-in relax).
- `OidcProviderService` + `OidcProviderController` (superadmin CRUD + discover-endpoints); client secret encrypted at rest.

Part of #7 · Depends on #17, #20

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

PR #798, opened 2026-08-31
feat(core): machine-credential authentication seam — decoder fallback in qnop-spi (#686)

### LINKED ISSUE #686: seam: let an extension contribute an authentication mechanism
## Why

Machine access to the API — service accounts for external systems — is being built as an enterprise extension (qnop-ee#19). The API itself already exists: 106 operations in the published OpenAPI contract (ADR-0015/0021). What it lacks is a caller that is not a person: the spec declares exactly one security scheme, `bearerAuth`, and that JWT belongs to a user.

**This issue is the core-side prerequisite** — making authentication extensible. Service accounts, keys, scopes and their admin surface stay in the enterprise repository.

## What stands in the way

`SecurityConfiguration` builds its chains in one `@Bean` method with every filter wired explicitly:

```java
.addFilterBefore(loginRateLimitFilter, CsrfFilter.class)
.addFilterBefore(changePasswordRateLimitFilter, AuthorizationFilter.class)
.addFilterAfter(passwordChangeRequiredFilter, AuthorizationFilter.class)
```

There is no point at which an extension can contribute. Everything an added mechanism would need — where in the order it sits, what it may set as the principal, how failures are rendered — is decided inside a method the extension cannot reach.

## A distinction worth making early

The core already runs **two** chains: the main one and an OIDC login chain scoped by `securityMatcher`. That means an extension can very likely add a *third* chain for its own paths today, without any core change — `@AutoConfiguration` plus `@Order` plus `securityMatcher("/api/ext/**")` is standard Spring.

If that holds, then **extensions authenticating their own endpoints need nothing from this issue.** The seam is only required for the harder case: a machine credential that authenticates against the **existing** API surface, which is exactly what qnop-ee#19 wants — a service account should be able to read reviews through the documented endpoints, not through a parallel set.

Confirming this is the first task here, because it decides how much seam is needed. It would be an easy mistake to design a broad authentication SPI when the actual requirement is narrow.

## The question to settle

**What may a contributed mechanism do, and what stays fixed?**

- **Principal shape.** A machine caller is not a `User`. Does the core grow a principal abstraction, or does an extension map its credential onto an existing identity — with all the problems a bot user has (qnop-ee#19 lists them)?
- **Authorisation.** Roles today are `ADMIN | MEMBER | AUDITOR`, all human. If a service account carries scopes instead, does the core learn scopes, or does the extension gate its own calls after the core has authenticated them? These give very different guarantees.
- **Ordering and precedence.** A contributed filter must not be able to place itself ahead of rate limiting or CSRF handling, and must not be able to weaken the existing chain. The seam has to constrain position, not just allow insertion.
- **Failure rendering.** A rejected machine credential should produce the same shape of error as any other auth failure, not the extension's own invention.

## Constraints

- **Nothing here may weaken human authentication.** A seam that lets an extension supply a principal is, by construction, a way to bypass the login path — it has to be impossible to use accidentally, and reviewable in isolation.
- **Rate limiting interacts.** Machine callers need limits keyed by their own identity, and #672 (buckets are process-local) hits harder on an endpoint designed to be called in a loop. Whether that is fixed here or separately, the seam should not make it worse.
- **Audit must stay attributable.** `AuditEvent.actorId` has to resolve for a contributed principal, so "who changed this" remains answerable when the answer is a machine.
- Same warning as the sibling seams: one consumer for now, so cut narrow and expect a re-cut (ADR-0049).

## Definition of done

- A written answer to whether a separate filter chain already suffices for extension-owned paths — with the finding recorded either way, because it determines this issue's scope.
- If a seam is needed: a decision recorded as an ADR, the contract published in `qnop-spi`, and constraints on ordering and principal shape enforced rather than documented.
- A test-only authentication contributor proving the seam works without an enterprise build, including that it cannot precede rate limiting.
- No change to existing authentication behaviour when nothing is contributed.

bigpuritz:
Closes #686 — second seam of epic #689. **Stacked on #797** (shares the EXTENSION-POINTS/contract-count edits); auto-retargets to `main` when #797 merges. Prerequisite for qnop-ee#19 (machine API access); keys, scopes, admin surface and machine-only endpoints stay enterprise.

## The scope question, answered first (as the issue demanded)

**Extension-owned endpoints need nothing from this issue.** The core already runs two `SecurityFilterChain` beans; a third chain scoped to `/api/ext/<id>/**` via `@AutoConfiguration` + `@Order` + `securityMatcher` is standard Spring with zero core change. Recorded in ADR-0060 and EXTENSION-POINTS §8. What remains is the hard case: a machine credential against the **existing** API surface.

## Decision (ADR-0060): a decoder fallback, not a filter seam

`io.qnop.spi.auth` is qnop-spi's fifth contract: `BearerCredentialAuthenticator` — `authenticate(credential) → Optional<MachinePrincipal>` (opaque non-UUID subject + extension-owned scopes). `DelegatingJwtDecoder` consults it **only after the local decode rejected the credential** and renders a claim as a synthetic JWT.

**The placement is the enforcement** — the issue's constraints hold structurally, not by convention:
- cannot precede rate limiting or CSRF (those run before bearer processing; there is no API to reorder anything)
- cannot observe or intercept valid user tokens (fallback fires only on local-decode failure; tested)
- cannot change failure rendering (unclaimed ⇒ the original rejection re-thrown; same 401 shape)
- cannot impersonate a user (UUID subjects rejected in code) and cannot mint human roles (scopes surface as `EXT_`-prefixed authorities only; the `role` claim of a machine principal is ignored) — so `requireUserId()` (all 80 call sites) answers 403 and every `hasRole` gate refuses machine principals

A machine principal is therefore *authenticated but nearly powerless* in the community core — deliberate: authorisation for machine access is the extension's business, and opening individual community read paths is the expected re-cut riding on the #684 principal work (ADR-0049 discipline).

Rides along: `AuthController` no longer 500s on a non-UUID subject at logout/me (previously unguarded `UUID.fromString`).

## Definition-of-done mapping

- [x] Written answer to the filter-chain question (ADR-0060 + EXTENSION-POINTS §8)
- [x] Seam published in `qnop-spi`; ordering/principal-shape constraints enforced in code, not documented
- [x] Test-only authenticator proves the seam incl. "cannot precede rate limiting" (structural: authenticators are only reachable from inside bearer decoding) and "cannot become ADMIN"
- [x] No change to existing authentication behaviour when nothing is contributed (fallback path unreachable without registered beans; valid-user-token test proves authenticators are never consulted)

Part of epic #689.


bigpuritz (commented):
## Independent agent review
Scope: full PR #798 diff relative to the #797 branch — `io.qnop.spi.auth`, `DelegatingJwtDecoder` fallback, `RoleJwtAuthenticationConverter`, `AuthController` hardening, ADR-0060, EXTENSION-POINTS §8 — verified against the security filter chain, `CurrentUser`, `PasswordChangeRequiredFilter`, the rate-limit filters and the error-rendering path.

**Verdict: Approve with comments.** The load-bearing security claims verify against the code. No CRITICAL or HIGH findings; two MEDIUM, four LOW.

### Claims I probed and confirmed

- **"Cannot precede rate limiting" is structurally true.** The IP-keyed limiters are `addFilterBefore(..., CsrfFilter.class)`; bearer processing (where the seam lives, inside `JwtDecoder`) runs in `BearerTokenAuthenticationFilter`, which Spring orders after `CsrfFilter`. There is no API through which an authenticator touches ordering.
- **Revoked user tokens never reach authenticators.** The try/catch wraps only `localDecoder.decode(token)`; the revocation check and the missing-claims rejections happen after it and their exceptions propagate without consulting the fallback. Combined with the `aValidUserTokenNeverReachesTheAuthenticators` test, the interception surface is exactly "credentials the local decoder rejected" — but see M1 for what that set contains.
- **The UUID-subject guard is coherent end-to-end.** `UUID.fromString` accepts uppercase/mixed-case, so case tricks are rejected; a whitespace-padded or otherwise non-canonical UUID-ish subject slips past `parsesAsUuid` but is then *also* unparseable by every consumer — I checked all of them: `CurrentUser.requireUserId`/`optionalUserId` (both `UUID.fromString` → 403/null, exactly as the javadoc promises), `AuthController.asUserId` (null), the converter (name passthrough), `ChangePasswordRateLimitFilter` (opaque string key, no parse). No consumer does a case-insensitive or substring comparison against user ids.
- **The role guard is airtight.** `EXT_` is *prepended*, so no scope can yield a `ROLE_*` authority (tested with a literal `ROLE_ADMIN` scope); `hasRole` prepends `ROLE_`, `CurrentUser.isAdmin()` matches `"ROLE_ADMIN"` exactly; `/actuator/**` and `/api/v1/admin/**` are role-gated. The `role` claim of a machine principal is never read.
- **Failure rendering is unchanged.** Unclaimed credentials re-throw the original exception, and even the guard's own `BadJwtException`s land in the entry point, which writes one fixed `UNAUTHENTICATED` body regardless of exception message — the 401 shape genuinely cannot vary.
- **`PasswordChangeRequiredFilter` claim accurate**: no `pcr` claim → `getClaimAsBoolean` null → `Boolean.TRUE.equals(null)` false → pass-through.
- **`AuthController` hardening changes nothing user-facing.** Real user subjects are always UUIDs, so `asUserId` is identity for them; the only behavioral change is non-UUID subjects going from unhandled 500 to defined 401/no-op. Logout with a machine bearer correctly skips the user-keyed revocation store.
- **SPI quality:** Optional-not-throw semantics, first-claim-wins bean order, constant-time note, and lifecycle ownership are all written into the contract; `MachinePrincipal` is immutable via `Set.copyOf`; the ArchUnit purity rule covers `io.qnop.spi.auth` automatically; fifth contract is a minor bump. ADR-0060 and EXTENSION-POINTS §8 match the code as built (the "80 call sites" figure checks out — 81 `requireUserId()` occurrences including the definition).

### MEDIUM

- **M1 — Expired genuine user tokens DO reach the authenticators; the "cannot observe valid user tokens" claim needs that caveat.** A signature-valid but expired (or clock-skew/nbf-rejected) user JWT fails the local decode and is then handed **verbatim** to every registered authenticator — real user token material: subject, role claim, jti, and an HMAC signature that is still valid for any verifier applying leeway. "Valid" in the PR/ADR claim quietly means "currently valid"; the observed set includes every token that *was* valid. Within the trusted-extension model this is not an exploit, but it is an information flow the ADR should state and the SPI javadoc should warn about: an authenticator will be handed expired user tokens and must treat unclaimed credentials as secrets — never log, store, or forward them. (If you want hardening beyond documentation, a cheap pre-filter — skip the fallback for credentials that parse as JWTs carrying the local token's characteristic claims — would shrink the exposure, at the cost of constraining extension credential formats to non-JWT-shaped or differently-shaped tokens; documenting may be the better trade.)

- **M2 — "Nearly powerless" is asserted but the residue is never enumerated or pinned.** A machine principal passes `.authenticated()`, so it reaches every community endpoint that neither requires a role nor resolves a user actor. That set is not empty: `BannerController` (banner read) is a concrete example, and any read method in user-facing controllers that doesn't call `requireUserId()` joins it. Today the residue looks harmless, but nothing *keeps* it harmless — a future `.authenticated()`-only read endpoint silently joins the machine-reachable surface with no test to notice. Recommend an integration test that authenticates a fake machine principal and asserts the expected answers across the API surface (or at least spot-pins the known-reachable endpoints), so the residue is a reviewed decision rather than an accumulating accident; plus one sentence in ADR-0060 naming the current residue.

### LOW

- **L1 — The converter keys on claim *presence*, not value.** `jwt.hasClaim(ACTOR_KIND_CLAIM)` treats any value as machine. Harmless today (user tokens have a fixed mint-time claim set; misclassification points in the fail-safe direction), but #21 plans layering OIDC decoders into this same class — check `"machine".equals(...)` for exactness before a third token source exists.
- **L2 — Dead code in `toJwt`:** `.claims(claims -> claims.putAll(Map.of()))` is a no-op — delete.
- **L3 — The rides-along `AuthController` fix is untested**, and its 401-vs-403 choice is quietly inconsistent: a machine principal hitting `/users/me/**` via `CurrentUser` gets 403 ("principal is not a user"), while `AuthController.requireAuthenticatedUserId` answers 401 ("Not authenticated") for the same principal. Both are defensible; a test for logout/me with a non-UUID subject plus a one-line comment on the divergence would finish the hardening.
- **L4 — Guard violations abort the whole fallback loop.** A buggy authenticator that *claims* a credential but returns a UUID/blank subject throws out of `authenticateMachine`, so later authenticators never get to claim — inconsistent with the throwing-authenticator case, which continues. HTTP-visible behavior is still a uniform 401, so this is resilience, not security: consider log-and-continue for guard violations too, or note that a claiming-but-invalid authenticator poisons the shared credential space.

### What is genuinely good
"The placement is the enforcement" is the right architecture for this seam — a decoder fallback gives extensions authentication without handing them a single ordering, rendering, or interception decision, and the rejected filter-contribution SPI in the ADR's alternatives is exactly the trap avoided. The smuggled-`ROLE_ADMIN`-scope test and the valid-token-never-consulted test cover the two attacks that matter most.

bigpuritz:
Review addressed in the follow-up commit:

- **M1 (expired tokens):** `JwtValidationException` is rethrown before the fallback — a signature-valid-but-expired user token never reaches authenticator code (unit-tested); ADR-0060 records the caveat, and the SPI javadoc obliges implementations to treat every offered credential as a secret (never log/store/forward).
- **M2 (surface pinning):** `MachineCredentialSurfaceIT` pins the residue — banner 200, documents 403, admin 403, audit 403, `/users/me` 401, unclaimed key = standard 401; ADR-0060 names the residue and that joining it is a deliberate decision.
- **L1:** converter checks the claim *value* (`"machine"`), not presence.
- **L2:** dead builder line removed.
- **L3:** 401-vs-403 divergence documented at `asUserId`; behaviourally covered by the IT.
- **L4 (rebutted, documented):** a claiming-but-guard-violating authenticator deliberately fails closed — letting the next authenticator claim a credential one already asserted as its own would be worse; reasoning is now a code comment at the call site.

bigpuritz (commented):
Scope: verification of the hardening commit (efa3634b, plus the build-scope fix a2c6c9c0) against my six findings on PR #798, with the requested fresh scrutiny on the `JwtValidationException` catch ordering.

**Verdict: five of six resolutions verified, the L4 rebuttal accepted — but the new surface-pinning IT asserts a status I believe is wrong, so one objection remains before this is clean.**

### The catch ordering — correct, and correct for the right reasons

`JwtValidationException` extends `BadJwtException` extends `JwtException`, so the subclass catch **must** come first and the compiler enforces it — an inverted order would not compile. More importantly, the *class choice* is right: the local decoder is a `NimbusJwtDecoder` (`JwtConfiguration`, `withSecretKey(...).macAlgorithm(HS256)`), whose default validators (`JwtTimestampValidator` et al.) throw `JwtValidationException` for expiry/nbf/skew, while parse failures and bad signatures throw plain `BadJwtException`. So exactly the dangerous set — signature-valid-but-expired genuine user tokens — is rethrown before the fallback, and exactly the harmless set (malformed or attacker-supplied material) still reaches authenticators. The `anExpiredUserTokenNeverReachesTheAuthenticators` test pins it with `verifyNoInteractions`. M1: **resolved**, and the ADR caveat plus the SPI "treat every offered credential as a secret — never log, store or forward" obligation are both in place.

### Also verified resolved

- **L1:** converter now requires `"machine".equals(getClaimAsString(...))` — value, not presence. Resolved.
- **L2:** dead builder line gone. Resolved.
- **L3:** the 401-vs-403 divergence is documented on `asUserId` with a defensible rationale. Resolved (but see the objection — the IT half of this item is the problem).
- **L4:** rebuttal accepted — the in-place comment ("a claimed principal violating an impersonation guard is a loud invariant breach; silently moving on would let another authenticator claim a credential one already asserted as its own") is a better argument than my suggestion; fail-closed is right here.
- **Build fix (a2c6c9c0):** moving `qnop-spi` to `qnop-app`'s main `implementation` scope is the correct resolution for the decoder's direct SPI use, and the comment's ArchUnit rationale is sound.

### Remaining objection — MEDIUM: the pinning IT asserts 401 for `/api/v1/users/me`, but that path should render 403

`GET /api/v1/users/me` is served by `CurrentUserController` (implements `UsersApi` — its own javadoc says "non-user (API-key) principals are rejected with 403"), not by `AuthController` as the IT's comment claims. The failure path is: `CurrentUser.requireUserId()` throws `AccessDeniedException` → `UnhandledExceptionHandler` deliberately rethrows it → `ExceptionTranslationFilter`, which for a **non-anonymous, fully authenticated** principal (a machine `JwtAuthenticationToken` is one) invokes the `accessDeniedHandler` → **403**. That is precisely the same machinery the IT itself relies on two lines earlier when it expects `/api/v1/documents` to answer 403 — the two endpoints take an identical path, so they cannot answer differently. `status().isUnauthorized()` on the `/users/me` line should be `isForbidden()`, and the "(401, see AuthController)" comment should move to an actual `AuthController` endpoint (`/api/v1/auth/**` is where `requireAuthenticatedUserId`'s 401 lives) or be corrected. Since Docker ITs only run in CI, this likely surfaces as a red CI run rather than a local failure — if CI instead passes as written, something in my model of the dispatch is wrong and I would genuinely want to know which mechanism turned it into a 401. Either way the test should assert the behavior the code actually has, with the comment naming the right controller.

Everything else is clean. Once the IT line is corrected (or CI proves it right and the comment is fixed to explain why), this PR has my approval.

bigpuritz:
IT assertion fixed: `/api/v1/users/me` → 403 (CurrentUserController via `requireUserId()`), comment corrected — AuthController's 401 divergence is scoped to `/api/v1/auth/**`. 🤖 Co-Author: Claude (Fable 5) via Claude Code

bigpuritz (commented):
Scope: verification of commit 2b1330e3 against the remaining objection from my second-pass review of PR #798.

**Verdict: resolved — Approve. All findings from both review passes are addressed; the #686 review loop is closed from my side.**

The IT now asserts `isForbidden()` for `GET /api/v1/users/me`, matching the actual dispatch (`CurrentUserController.requireUserId()` → `AccessDeniedException` → 403, the same path as every user-actor endpoint), and the corrected comment confines `AuthController`'s 401 divergence to `/api/v1/auth/**` — which is precisely where it lives. With CI green including the Docker IT, the pinning test now proves what it claims: the machine-reachable residue is the banner read, user-actor endpoints answer 403, role gates answer 403, and an unclaimed credential stays the standard 401.

Final state of the seam as reviewed: expired user tokens never reach third-party authenticators (rethrow before fallback, tested), the two impersonation guards are airtight end-to-end, ordering/rendering/rate-limit precedence are structural rather than conventional, the residual surface is pinned by an integration test, and the contract documents its obligations (secret handling, constant-time comparison, Optional semantics). This is a carefully built security seam.

Issue #602, opened 2026-07-24
feat(review): extension seams for the live review feed — event projection, audience/identity facade, live-channel UI slot


bigpuritz:
## Motivation

The live annotation/comment feed over SSE is an enterprise feature (first consumer: qnophq/qnop-ee#5). Its design deliberately rides on machinery the community already has — the committed `ReviewEvent` stream, the notification path's recipient resolution, per-review anonymity (ADR-0038) and the feed's query keys. But `qnop-ee` builds against the published `qnop-spi` artifact only (ADR-0002/0003): none of those internals are reachable from an enterprise module today. This issue adds the seams — and no live-feed code itself.

## Scope — server

1. **Review-event visibility.** `io.qnop.service.review.ReviewEvent` (sealed: `AnnotationCreated`, `CommentAdded`, `AnnotationDecided`, `WorkflowChanged`, `ParticipantAdded`, `VersionUploaded`) is internal. Expose a stable, versioned projection of these events to extensions — either by promoting the event contract into `qnop-spi` or via an extension-consumable re-publication — so an enterprise listener can subscribe **after commit** without compiling against `qnop-core`. Decide the vehicle deliberately (ADR if `qnop-spi` grows; consistent with the same question in #598).
2. **Recipient + identity facade.** The scoping the e-mail path already trusts must be reusable, not re-implemented: who belongs to an event's audience (review circle / thread members, actor excluded — `ReviewNotificationService`'s recipient resolution) and under which display identity per the review's anonymity policy (`ReviewIdentityResolver`, ADR-0038). Expose a narrow read-only facade for extensions; the e-mail path keeps using the internals directly.
3. **Access-check facade.** The SSE endpoint itself is an enterprise controller, but it must authorize with the *same* rule as `listAnnotations` (owner/participant visibility). Make that check callable through the facade so the two rules cannot drift.

## Scope — UI (runtime UI extension model, ADR-0039)

- **Review live-channel extension point:** a registered extension gets a lifecycle hook while a review surface is mounted (`documentId` in, teardown on unmount) — the place an SSE client would live.
- **Stable invalidation API:** the extension must be able to say "annotations of document X changed" / "comments of annotation Y changed" without importing `qnop-ui`'s internal query keys. Expose a small invalidation facade over the existing keys (`annotationKeys`, `commentKeys`), so a push event translates into the normal authorized refetch.

## Deliberately NOT in scope

[...]

Issue #536, opened 2026-07-20
feat(audit): free-text search over the audit event details


bigpuritz:
The audit trail (#466, ADR-0042) can filter by event type, actor, document, and time range — but not by content. Operators need a free-text search over the **Details** column, e.g. to find every event mentioning a storage key, a job id, or a workflow state.

Scope (server-side — the table is paginated, so client-side filtering would only search the current page):

- OpenAPI: new optional `detail` query param on `GET /audit/events` — case-insensitive substring match on the raw detail JSON; LIKE wildcards in the input match literally
- `AuditLogService`: one additional Specification predicate (`lower(cast(detail as text)) LIKE %…%` with escaped wildcards); the `detail` column is jsonb, so the cast is required
- `AuditController`: pass-through
- qnop-ui: debounced search field in the audit filter row, wired through `useAuditLog` (pagination resets on change)

🤖 Co-Author: Claude (Fable 5) via Claude Code

PR #799, opened 2026-08-31
feat(core): account-less review participants — ExternalParticipants facade in qnop-spi (#684)

### LINKED ISSUE #684: seam: a review participant without an account

bigpuritz:
Closes #684 — the third and largest seam of epic #689. **Stacked on #798** (auto-retargets on merge). Prerequisite for qnop-ee#17 (guest links); credentials, invitations, expiry and UI stay enterprise.

## The principal decision (ADR-0061, shared with #686 as the epic demanded)

ADR-0060 fixed the *authentication* shape (subject string; UUID = user). ADR-0061 fixes the *review-domain* shape: **every actor is a UUID principal id** — a user's is their user id; an account-less participant is a `review_participant` row (`user_id`/`team_id` null, `external_display_name` set, three-way XOR CHECK) whose **own row id is the principal id**, flowing through authorship, access, visits and anonymity ordinals without any of them learning a second concept.

## The seam: qnop-spi's sixth contract — and its first core-implemented facade

`io.qnop.spi.participant.ExternalParticipants` (`add` → principal UUID, `remove`, `hasAccess`): the extension calls the core (the direction #602's facades will repeat; new row in the EXTENSION-POINTS conventions table). Invariants stay behind it:

- `remove` refuses user/team rows — an extension can never manage the account-bearing roster
- validation + audit (null actor, `external: <name>` detail) core-side; **no `ReviewEvent`** for guest roster changes — whom to notify about a guest is the extension's decision
- **Anonymity (ADR-0038):** guests number into the same ordinal space; in an anonymous review a guest reads as "Participant N" — a supplied display name cannot pierce it (IT-proven)

## Mechanics

- `existsAccessibleParticipant` gains a third leg (`p.id = :principal AND external`) — `DocumentAccessService` and all callers unchanged
- `review_visit.user_id` drops its FK to `qnop_user` (holds principal ids now); folded into changeset 0011 per the fold rule, like the new column and CHECK
- `ParticipantKind` gains `EXTERNAL` (additive; FE consumers compare `=== Team`/`=== User` and degrade safely — verified); the workspace principal directory deliberately never lists external participants
- Guest authorship endpoints, guest-scoped `thread_participation`, and widened audit actor semantics are deliberately deferred until qnop-ee#17 forces them (ADR-0049)

## Definition-of-done mapping

- [x] Boundary decision recorded (ADR-0061: what the core knows / refuses to learn)
- [x] Seam published in `qnop-spi` (purity ArchUnit-covered)
- [x] Test-only consumer: `ExternalParticipantServiceTest` (facade guards) + `ExternalParticipantIT` (schema, access leg, EXTERNAL listing, principal-keyed visits, anonymous pseudonym)
- [x] Behaviour unchanged for account-bearing rosters (existing tests untouched; view mapping keeps USER/TEAM byte-identical)

## Test plan
- [x] Unit tests + ArchUnit + `:qnop-spi:build` + Spotless green; FE `tsc` + suite green (isolated; full-run flakes are the known load-timeout pattern, CI authoritative)
- [ ] CI green (Docker ITs incl. the new `ExternalParticipantIT`)

Part of epic #689.


bigpuritz:
CI caught a real one: `audit_event.detail` is a JSON column — the facade's audit rows now write `{"participantId":"…","kind":"external"}` (identifiers only; the display name is user content and stays out). 🤖 Co-Author: Claude (Fable 5) via Claude Code

bigpuritz (commented):
## Independent agent review
Scope: full PR #799 diff relative to the #798 branch — `io.qnop.spi.participant`, `ExternalParticipantService`, the schema fold, identity/roster/access changes, ADR-0061, EXTENSION-POINTS §9 — verified against `ReviewIdentityResolver`, `ReviewParticipantService`, `DashboardService`, `AuditLogService`, the 0011/0016 changesets, `AdminUserService`, and the frontend `ParticipantKind` consumers.

**Verdict: Approve with comments — after two fixes.** The design is right: one principal-id space, a core-implemented facade with the invariants genuinely behind it, and the anonymity *mechanism* for guests is sound. But two HIGH findings need fixing (both would likely surface as red CI, one of them in the very IT that claims to prove the anonymity story), plus one MEDIUM and four LOW.

### HIGH

- **H1 — The audit detail is plain text into a `jsonb` column; the insert should fail at runtime.** `audit_event.detail` is `JSONB` (0011 changeset: `- column: { name: detail, type: JSONB }`), the entity maps it `@JdbcTypeCode(SqlTypes.JSON)`, its javadoc says "optional **jsonb** context", every existing writer stores JSON objects (`{"annotationId":"…"}`), and the #536 free-text filter builds a text projection *of jsonb*. `ExternalParticipantService` writes `"external: " + name` — not JSON — so Postgres should reject the value (`invalid input syntax for type json`), aborting the extension's `add()`/`remove()` transaction. The unit test can't see this (mocked repository); `ExternalParticipantIT.add(...)` against the real schema should trip it at auto-flush. Fix: emit JSON, e.g. `{"external":"<name>"}` (escaped — display names may contain quotes/backslashes; reuse the existing `escape(...)` style in `ReviewDeletionService` or a proper writer), which also keeps the audit UI's detail filter working.

- **H2 — The anonymity IT asserts the owner's view — the one viewer ADR-0038 exempts.** `anAnonymousReviewPseudonymizesTheGuestLikeAnyReviewer` calls `participantService.list(documentId, owner, false)` where `owner` IS the document owner. `list()` computes `anonymise = anonymous && !admin && !ownerId.equals(actor)` → **false** for the owner → the non-anonymised branch returns the guest's real `"Guest Reviewer"` and real principal id — both assertions (`startsWith("Participant ")`, `principalId != principal`) fail, and per ADR-0038's #422 amendment that owner-sees-real-roster behavior is *correct*. So the test contradicts both the code and the ADR, will fail CI, and the PR's "IT-proven" anonymity claim is unproven as written. The fix is to view as a non-owner: add a second (user) participant and call `list(documentId, thatParticipant, false)` — then the pseudonym and synthetic-token assertions are the right ones. To be clear about the underlying mechanism, which I verified independently and is sound: the resolver adds external row ids into the ordinal set *before* ordinal assignment (same sorted space), `displayName`/`exposedAuthorId`/`slug` treat the guest key exactly like a user id (pseudonym + synthetic token + null slug for non-owner viewers), `anonymiseRoster`'s external branch keys correctly by row id, and the guest's real name reaches only the ADMIN/AUDITOR audit surface (sanctioned unmasking, ADR-0042) — the participant-facing dashboard feed neither includes `review.participant_added` in `FEED_EVENT_TYPES` nor ships `detail` at all. The proof just tests the wrong actor.

### MEDIUM

- **M1 — Dropping `fk_review_visit_user` silently loses the user-deletion cleanup.** Users *are* hard-deleted (`AdminUserService.delete`, whose javadoc enumerates what "the database cascades"); until now `ON DELETE CASCADE` removed the user's `review_visit` rows, and after this PR they linger as orphans until the document dies. The changeset comment addresses only principals removed *from a review*, not deleted users. Harmless operationally (the rows are keyed reads by the caller), but it is a behavior regression made implicitly — either add the one-line cleanup to `AdminUserService.delete` or extend the changeset comment to state that deleted users' visit rows are deliberately left to document-cascade. A decision, not an accident.

### LOW

- **L1 — `hasAccess` answers more than its javadoc claims.** It delegates to the general `existsAccessibleParticipant`, so a user id or team member's id also returns `true` — the facade javadoc ("whether the **external** principal participates") undersells it. Either document the one-principal-space semantics ("any principal") or constrain the query to external rows; the current mismatch will confuse the first extension author who probes it.
- **L2 — Guests' principal ids are fed to the user-avatar endpoint.** `ReviewHubHead`'s non-team branch calls `avatarSrc(participant.principalId)` for EXTERNAL rows → `GET /api/v1/users/<row-id>/avatar` → 404 → initials fallback. "Degrades safely" holds, but it's a wasted request per guest per render and cuts against the catalogue's own "a contributed id must never be fed to user endpoints" sensibility (§6). A `kind`-aware guard belongs with the qnop-ee#17 UI work at latest.
- **L3 — Display-name validation is trim+length only.** Control characters, zero-width and bidi-override codepoints are stored verbatim and rendered in the roster and the admin audit trail (React escapes HTML, so this is a spoofing surface, not XSS). Rejecting ISO control chars (and perhaps normalizing) is cheap hardening for a string an extension relays from the outside world.
- **L4 — ADR-0061 overstates authorship today.** "The row's own id … flowing through authorship columns" — `annotation.author_id` and `comment.author_id` still carry `fk_annotation_author`/`fk_comment_author` → `qnop_user`, so a guest structurally *cannot* author anything yet (which incidentally makes the guest-authorship deferral safely enforced, not just declared — good). One clause in the ADR noting that the authorship re-cut must widen those FKs (as `review_visit` did here) would keep the "one principal-id space" claim honest about its current reach.

### What I verified and found sound
The three-way `num_nonnulls` XOR is exactly right (blocks user+external combos; external rows carry only the name). The third access leg is confusion-free: `p.id = :userId` is guarded by `externalDisplayName IS NOT NULL`, UUID collision between row ids and user ids is not a practical concern, and every other consumer of a principal id either parses/looks it up as a user (misses harmlessly) or goes through this one query. `remove()` leaks nothing across documents (the document filter runs before the external check, so a foreign row answers plain `false`). `ParticipantKind.EXTERNAL` is additive on a string enum; the FE's `=== Team`/`=== User` comparisons keep guests out of mention rosters and the principal directory correctly. No `ReviewEvent` for guest roster changes is the right call — it also keeps #797's published stream honest, since no `review.participant.added` fires for a principal the extension alone knows how to describe. EXTENSION-POINTS §9 and the new conventions-table row match the code; the core-implemented-facade direction is well chosen and well explained.

bigpuritz:
Review addressed in the follow-up commit (the JSON-detail HIGH was already fixed in the previous one — identifiers-only rather than escaped name, deliberately):

- **H2:** the anonymity IT now looks through a **non-owner participant's** eyes (owner exempt per ADR-0038 #422); assertions unchanged and now meaningful.
- **M1:** recorded as an accepted consequence (changeset comment + ADR-0061): orphaned visit rows of hard-deleted users are inert — readers join names by id and skip unknowns — and cascade away with the document.
- **L1:** `hasAccess` javadoc now states it answers the core's one access rule for any participating principal.
- **L2:** `ParticipantsDialog` renders initials for EXTERNAL principals instead of querying the user-avatar endpoint.
- **L3:** display names reject control, zero-width and bidi-override characters.
- **L4:** ADR-0061 now names the authorship-FK reality (`annotation/comment.author_id` still FK `qnop_user`) as the enforcing mechanism of the deferral and the re-cut's job.

bigpuritz (commented):
Scope: verification of commits 14ee3f5b and 34212691 against my seven findings on PR #799.

**Verdict: Approve. All findings addressed; one cosmetic residual noted, non-blocking. This closes my review of the epic #689 seams.**

- **H1 (jsonb audit detail): resolved — and better than asked.** `{"participantId":"<uuid>","kind":"external"}` is valid JSON of pure identifiers, which sidesteps the escaping question entirely instead of answering it; the deliberate trade (the name is readable from the row while it exists, and after removal the trail names only the id) is stated in the code comment and consistent with the identifiers-only philosophy the epic has held throughout.
- **H2 (anonymity IT viewed as owner): resolved.** The IT now adds a fellow user participant and asserts through their eyes — the one perspective where ADR-0038's pseudonym rule actually applies — and CI is green including this IT, so the "guest pseudonymizes like a user" claim is now genuinely proven: "Participant N", synthetic principal id, real name withheld.
- **M1 (visit-orphan regression): resolved as a recorded decision.** The changeset comment now names hard-deleted users explicitly and argues the rows are inert (readers join by id and skip unknowns; document-cascade bounds them), and ADR-0061's consequences carry the same. A decision, not an accident — exactly what was asked.
- **L1: resolved.** `hasAccess` javadoc now states it answers the core's one access rule for any participating principal, with guidance for guest-credential gating.
- **L2: resolved in `ParticipantsDialog`; one spot remains.** The dialog now short-circuits EXTERNAL to initials. `ReviewHubHead`'s stacked-avatar row (~line 317) still calls `avatarSrc(participant.principalId)` in its non-team branch, so a guest in the hub head still costs one 404 per render. Same LOW as before, degrades safely, and the qnop-ee#17 UI work is the natural place to sweep it — flagging only so it is known, not to hold the PR.
- **L3: resolved.** ISO controls, U+200B–200F, the bidi embeddings/overrides and isolates are rejected — the spoofing set that matters is covered.
- **L4: resolved.** ADR-0061 now states the authorship-FK reality (guests structurally cannot author until the re-cut widens `fk_annotation_author`/`fk_comment_author` deliberately) and the identifier-only audit detail.

Closing note on the epic: across #797/#798/#799 the same shape held up under review three times — invariants core-side, placement as enforcement, test-only consumers proving the seam, and a catalogue that stayed in sync in the same PR. The findings that mattered (anonymity crossing seams, expired-token exposure, wrong-viewer tests) were all of one family — the gap between a stated invariant and the code path that proves it — and every one was closed with the invariant strengthened rather than the claim softened. Good epic.

bigpuritz:
**Correction (maintainer catch):** the schema changes were folded into shipped changeset 0011 — illegal now that qnop has published releases (Liquibase checksum validation would fail on every deployed instance). Fixed:

- `0011-document-review-schema.yaml` is byte-identical to `main` again.
- The column, the widened three-way principal CHECK (drop+add via `sql` — CHECKs have no typed change) and the `review_visit` FK drop live in the **new `0037-external-participants.yaml`**, included after 0036 and before the enterprise `includeAll`.
- The retired fold practice is replaced by **CLAUDE.md working rule 11**: released changesets are immutable; every schema change is a new migration.

bigpuritz (commented):
## Independent agent review — migration-correction verification
Scope: verification of commit eb6960b0 (released-changeset immutability correction) on PR #799.

**Verdict: verified correct on all four points — Approve stands.**

- **0011 is truly byte-identical to main.** Not taken from the patch: I fetched `0011-document-review-schema.yaml` at `main` and at the PR head and diffed the full contents — zero differences.
- **0037 is semantically equivalent to what was reviewed.** Same column (`external_display_name VARCHAR(120)`, matching `MAX_DISPLAY_NAME = 120`), same three-way `num_nonnulls(user_id, team_id, external_display_name) = 1` expression, same `fk_review_visit_user` drop with the accepted-orphan decision comment carried along. The partial unique indexes (`ux_review_participant_user`/`_team`) correctly need no change — both are keyed on non-null user/team and external rows have neither.
- **The CHECK drop+add is sound in both lineages.** A deployed instance has the two-way constraint from shipped 0011, so the unconditional `DROP CONSTRAINT` finds it; a fresh install runs 0011 (two-way) then 0037 (widen) in order and ends identical. `splitStatements: true` is set for the two statements, and using a `sql` change is right — CHECK constraints have no typed Liquibase change. The rationale itself is correct: qnop has published releases, and Liquibase's checksum validation would have failed on every deployed instance the moment edited-0011 shipped — this correction genuinely averts a production breakage, and working rule 11 turns the lesson into policy.
- **Include position is correct.** `0037-external-participants.yaml` sits directly after 0036 and before the enterprise `includeAll` (master changelog lines 120/123/131), so enterprise changesets still run after all community migrations.

**One LOW follow-up, non-blocking:** with the fold rule retired by working rule 11, `docs/EXTENSION-POINTS.md`'s Liquibase-seam entry still says 'The community fold-rule ("modify the existing changeset") does **not** apply across the boundary' — a rule that no longer exists on either side of the boundary. One-line update whenever that file is next touched (the historical "amended into the existing 0011 changeset, pre-prod convention" wording inside old ADRs is fine as history).

For the record, and applicable to my own prior reviews too: the original fold-into-0011 change passed my review because I checked it against the project's then-documented convention ("schema changes fold into existing changeset, not yet in production") rather than against the release reality. With releases published, that convention was stale — rule 11 now makes the correct check trivial: any diff inside an existing `migrations/NNNN-*.yaml` is a defect, full stop.

bigpuritz:
LOW addressed: the EXTENSION-POINTS Liquibase-seam entry now states rule 11 (shipped changesets immutable on both sides of the boundary) instead of the retired fold rule. 🤖 Co-Author: Claude (Fable 5) via Claude Code

PR #804, opened 2026-08-31
feat(core): land the approved auth + external-participant seams on main (#686, #684)

### LINKED ISSUE #684: seam: a review participant without an account

### LINKED ISSUE #686: seam: let an extension contribute an authentication mechanism

bigpuritz:
Closes #686, closes #684.

**Why this PR exists:** the stacked chain was merged bottom-up, but #798 was squashed into `feat/issue-685-event-seam` and #799 into `feat/issue-686-auth-seam` *before* GitHub retargeted them — so only #797's content reached `main`; the auth seam (ADR-0060), the external-participant seam (ADR-0061) and migration 0037 were stranded on the (now dead) feature branches.

This PR cherry-picks the two squash commits onto current `main`, **content byte-identical to the finally-approved, CI-green PR heads** (`f941f08d` = #798, `5d592f86` = #799 — both carry their full review history in the PR threads). No new code; the independent reviews and their fixes all apply verbatim.

## Test plan
- [x] Local: `:qnop-spi:build`, facade + machine-auth unit tests, ArchUnit green
- [ ] CI green (Docker ITs incl. `MachineCredentialSurfaceIT` and `ExternalParticipantIT` against migration 0037)

Part of epic #689 — after this merges, the epic closes.


bigpuritz (commented):
## Independent agent review — landing verification
Scope: PR #804 only as a re-landing vehicle — verifying its tree is content-identical to the finally-approved seam work; no re-review of the content (it carries my final approvals from #798 and #799).

**Verdict: confirmed — the tree is content-identical to the approved head outside main-side dependency churn. Approve.**

- `git diff origin/feat/issue-686-auth-seam <PR-head 989743b2> --name-only` yields exactly four files: `docs/branding/package.json`, `gradle/libs.versions.toml`, `qnop-ui/package.json`, `qnop-ui/pnpm-lock.yaml` — nothing else in the tree differs from the finally-approved branch head (5d592f86, itself one of the two cherry-picked squash commits).
- Every one of those diffs is attributable to Renovate merges on main that post-date the branch: the npm bumps (`@tanstack/react-query` 5.102.8, `axios` 1.20.0, `lucide-react` 1.35.0, `@testing-library/react` 16.3.3, `@vitejs/plugin-react` 6.1.1, `sharp` 0.35.4 + lockfile) are precisely commit bdcb1dab (#803), and the toml bumps (`swagger-annotations` 2.2.54, `awssdk` 2.54.6) are #801/#802 (b6b87541, d932c19b).
- Direction check: for all four files the PR head is byte-identical to `origin/main` — the cherry-pick correctly kept main's newer versions; no seam content was lost and no stale dependency was reintroduced.

So #804 = approved #798+#799 content, rebased onto current main via the two squash cherry-picks, with zero semantic delta. Good catch on the stacked-merge mishap — this is the right way to land it.

PR #807, opened 2026-08-31
feat(review): review facades + live-channel UI slot — the last live-feed seams (#602)

### LINKED ISSUE #602: feat(review): extension seams for the live review feed — event projection, audience/identity facade, live-channel UI slot

bigpuritz:
Closes #602 — the last seam of the enterprise wave. Prerequisite for qnop-ee#5 (live feed); the SSE endpoint, emitter registry, heartbeats, reconnect and any broker (ADR-0013) stay enterprise, and the e-mail path keeps consuming internal events unchanged.

## Scope note

The issue's first point — a published event projection — was settled by ADR-0059/#797 (`PublishedEventListener`); this PR delivers the remaining three seams (ADR-0062):

## Server — `ReviewFacade`, qnop-spi's seventh contract (core-implemented)

- **`mayView(documentId, principalId)`** — exactly the annotation-listing visibility rule; admin override deliberately absent. Covers ADR-0061 account-less principals via the third access leg. **IT-pinned** against the listing service for owner, direct participant, team member and outsider (`ReviewFacadeIT`).
- **`reviewCircle(documentId)`** — the notification audience. The computation moved into `ReviewFacadeService.circleOf` and `ReviewNotificationService` now delegates — **one implementation, no drift possible**. Account-less participants deliberately excluded (no notification identity; `ExternalParticipants.hasAccess` admits them); the actor is not pre-excluded (consumers subtract the published event's `actorId`).
- **`displayNameFor` / `exposedAuthorIdFor`** — the ADR-0038 per-viewer identity; what leaves the process must be these values, never raw ids.

## UI — live-channel slot + invalidation facade

`extensions/liveChannel.ts`: a `LiveChannelContributor.onReviewMounted(context)` fires while a review surface is mounted (re-fires per review navigation) and returns its teardown; the context carries `documentId`, `invalidateAnnotations()`, `invalidateComments(annotationId)` — closing over the host's query client and keys, so a push becomes a normal authorized refetch without the extension importing internals. A throwing contributor costs itself, never the page. `DocumentReviewPage` mounts the hook — byte-identical without registrations. Registered per-feature; migrates to the generic slot registry (#600) at the `qnop-ui-spi` cut.

## Acceptance criteria

- [x] Community behaviour unchanged without extensions (no-op hook; notification tests untouched and green through the shared-circle refactor)
- [x] Test-only fakes prove: committed events arrive (#797's suite), audience/identity resolve through the facade (`ReviewFacadeServiceTest`), invalidation triggers the normal refetch (`liveChannel.test`)
- [x] Access facade answers exactly like the listing authorization for all four roles (`ReviewFacadeIT`)
- [x] ADR-0062; EXTENSION-POINTS §11/§12 + planned rows closed (same-PR rule)

## Test plan
- [x] Unit + ArchUnit + Spotless + `:qnop-spi:build` green; FE `tsc`/`eslint`/`prettier` + liveChannel & DocumentReviewPage suites green
- [ ] CI green (Docker ITs incl. `ReviewFacadeIT`)


bigpuritz (commented):
## Independent agent review
Scope: full PR #807 diff against `main` — the `ReviewFacade` contract and `ReviewFacadeService`, the `ReviewNotificationService` refactor, `ReviewFacadeIT`, the live-channel UI slot, ADR-0062 and EXTENSION-POINTS §11/§12 — verified against `DocumentAccessService`, `AnnotationService.list`, `ReviewIdentityResolver`, the notification pipeline and the FE query-key factories.

**Verdict: Approve with comments.** No CRITICAL or HIGH (and no `migrations/NNNN-*.yaml` touched); three MEDIUM, four LOW.

### The stress points that verify clean

- **The mayView pinning is genuinely one rule, not two that agree today.** `mayView` → `access.isVisible(…, false)` → the private `canAccess`; the listing → `getDocument` → `requireVisible` → the *same* `canAccess`. The IT's owner/direct/team-member/outsider matrix therefore pins two doors of one predicate — drift would require editing `canAccess` itself, which changes both callers. The ADR-0061 third access leg rides along via `existsAccessibleParticipant`, so the "covers account-less principals" claim is true. No-admin-override is sound and stated in the contract.
- **The circle refactor is behaviour-preserving by construction.** `circleOf` is the old loop moved verbatim (same repositories, same LinkedHashSet order, same inputs — the notification service feeds its already-loaded owner id); the delegation is one line; and `ReviewNotificationServiceTest` wires a *real* `ReviewFacadeService` over the same mocks, so the existing notification tests still exercise the actual circle computation rather than a stub. Account-less exclusion is tested; actor-not-pre-excluded is documented in the SPI, the ADR and §11 consistently.
- **Identity answers cannot leak what ADR-0038 hides.** `displayNameFor`/`exposedAuthorIdFor` delegate to `ReviewIdentityResolver.forDocument(documentId, viewerId)` — the per-viewer resolution whose exposure rules (self/owner exempt, pseudonym + synthetic token otherwise, guests numbered into the same ordinal space) I verified in the #684 review. The "what leaves the process must be these values" obligation is written into the contract.
- **The UI slot's lifecycle is right and tested:** remount per review navigation, teardown on unmount, a throwing contributor costs itself (and a throwing teardown too), byte-identical without registrations (early return, no context even built). The invalidation shapes are correct: `commentKeys.list(annotationId)` uses the factory verbatim, and `[...annotationKeys.all, 'list', documentId]` is a true prefix of `annotationKeys.list(documentId, version)` — so "every cached version window" is accurate.

### MEDIUM

- **M1 — The facade omits the thread-participation half of "the listAnnotations authorization rule", silently.** The planned-row text this PR closes promised the facade would cover "the `listAnnotations` authorization rule" — but that rule has two layers: the document gate (`canAccess`, delivered) and the per-thread PRIVATE filter (`canSeeThread`, absent). In a `PRIVATE` review, a participant passes `mayView` yet may not see foreign threads — and the published events (#797) carry `annotationId`/`commentId`, so an event-granular live feed fanned out by `mayView` + `reviewCircle` delivers thread-existence and activity metadata to principals the API deliberately 404s (the listing filters per-thread; the anti-enumeration comment in `requireVisible` shows how seriously the core takes this). To be fair: the mail path is equally thread-blind (I checked — `ReviewNotificationService` never consults the policy), so the facade widens nothing the core doesn't already do by mail; that is why this is MEDIUM, not HIGH. But the seam's own charter is "reuse, never re-implement, the security-relevant scoping", and the first consumer has no facade-supported way to be thread-correct. Either add the thread predicate (`mayViewThread(documentId, principalId, annotationAuthorId)` — the `canSeeThread` inputs) or state in the contract, §11 and ADR-0062 that per-thread PRIVATE scoping is out of contract and event-granular consumers must stay coarse ("something changed — refetch"), which the invalidation facade happens to make possible. Today the contract is simply silent on `thread_participation`, and silence is how the next leak gets built in good faith.
- **M2 — `useLiveChannel` misses contributors that register after mount — the loader's normal timing.** The hook reads the module-level `contributors` array inside an effect whose deps are `[documentId, queryClient]`: no subscription, no re-run on registration. #805's generic registry used `useSyncExternalStore` for precisely this reason (its own test: "the runtime loader resolves its dynamic imports AFTER the first render"). As shipped, an ADR-0039-loaded SSE extension that registers after `DocumentReviewPage` mounted attaches only when the user navigates to another review — a silently dead live channel on the first page view. Subscribe the hook (the mentions registry's notify pattern is already in this directory), or make the §12/ADR "migrates to the generic slot registry" note carry the known limitation explicitly until the migration happens.
- **M3 — The catalogue on this branch references a section it does not contain.** This branch's EXTENSION-POINTS goes §9 → §11/§12 — §10 is reserved for #805's message-row entry, and #805 (like #806) is still open; §12's text says "Migrates to the generic slot registry (§10)…" pointing at a section that does not exist here. Three open PRs are editing the same file and the same planned-rows table; whichever merges first, the others conflict, and if #807 lands first, main ships a numbering gap and a dangling cross-reference. Needs a deliberate merge order (or renumbering on rebase) — flagging so it is planned, not discovered.

### LOW

- **L1 — `reviewCircle` includes disabled users, and the javadoc implies otherwise.** The mail path filters `User::isEnabled` (and the actor) *after* the circle; the SPI text "the review circle the notification path addresses" reads like the final audience. One sentence — enabled-ness and opt-out filtering stay with the consumer — keeps the first SSE author from mailing-list-style surprises.
- **L2 — Edge behavior is inconsistent across the facade.** `mayView(unknownDoc, …)` returns false, but `displayNameFor`/`exposedAuthorIdFor` throw `DocumentNotFoundException` for the same id, and `displayNameFor(doc, viewer, null)` NPEs (`authorId.equals` inside the resolver). Trusted callers, but a read-only facade should state its null/unknown contract — or answer benignly everywhere.
- **L3 — The annotation invalidation hand-builds the factory's internal layout.** `[...annotationKeys.all, 'list', documentId]` duplicates the prefix of `annotationKeys.list()`; a future key-shape refactor breaks the facade silently (the test asserts the same hand-built shape, so it would keep passing). Export a `listPrefix(documentId)` from the factory and use it in both places — that closes the drift the facade exists to prevent.
- **L4 — Small finishers:** ADR-0062's consequences line contains literal `\'` escapes ("The facade\'s answers … endpoint\'s behaviour") — an escaping artifact to clean; and the IT could cheaply add the two cases its javadoc-level claims rest on (an account-less principal through `mayView`, and one anonymous-review `displayNameFor`/`exposedAuthorIdFor` assertion at IT level — currently unit-mocked only).

Contract counts (seven) are updated consistently in CLAUDE.md and ARCHITECTURE.md; `io.qnop.spi.review` is automatically covered by the ArchUnit purity rule; the rejected alternatives in ADR-0062 (no repository exposure, no pre-resolved audiences) are the right calls with the right reasons. The facade-direction pattern from ADR-0061 is wearing in well.

bigpuritz:
Review addressed (staged-file list verified before this comment — the mapping below is written from the diff):

- **M1:** the contract states **review-level only** — `mayView` javadoc, ADR-0062 and §11 all carry the PRIVATE-thread caveat and the identifier-only obligation for event-granular fan-out; a thread-level facade waits for a consumer (ADR-0049). The mail path's equal blindness is noted as the parity argument.
- **M2:** `useLiveChannel` subscribes via `useSyncExternalStore` — a contributor registered after mount (the ADR-0039 loader timing) reaches an already-open review; teardown/remount covered by the existing tests.
- **M3:** §12 now names the #805/#807 collision explicitly (whichever merges second reconciles the numbers) instead of referencing a not-yet-existing §10.
- **LOW:** circle javadoc states disabled accounts are included (filtering is the consumer's, as the mail path's); identity methods answer degraded values on null edges like `mayView`; `annotationKeys.listPrefix` exported and used (no hand-built key shapes); ADR escapes fixed; IT gains the account-less `mayView` case and an order-insensitive circle assertion.
- **CI red herring:** the failing run predates the seed-collision fix (its log still shows `username=member`); seeded identities are now fully unique including display names.

bigpuritz (commented):
Scope: verification of commits 87c18e82 and 359a5eea against my seven findings on PR #807; each checked in the commit diff AND against the head-state files.

**Verdict: six of seven resolved cleanly — one small item remains: the M1 caveat landed in the contract and the ADR but not in §11 of the catalogue.**

### Verified resolved

- **M2 (late registration): resolved, and correctly.** `useLiveChannel` now subscribes via `useSyncExternalStore` with `registered` in the effect deps — a loader-registered contributor reaches an already-open review immediately (head-confirmed). Note in passing: a new registration remounts *all* contributors for the surface (teardown + reattach); inherent to the effect model, harmless for SSE, no action needed.
- **M3 (§-collision): resolved.** §12 now names the #805/#807 reconciliation plan explicitly ("whichever merges second reconciles the section numbers") instead of dangling a cross-reference.
- **L1: resolved** — `reviewCircle`'s javadoc now says "INCLUDING disabled accounts" with the consumer obligation stated.
- **L2: resolved for the null edges** — `displayNameFor`/`exposedAuthorIdFor` now answer their degraded values, mirroring `mayView`. The unknown-documentId asymmetry (identity methods still throw where `mayView` returns false) remains, and the commit honestly doesn't claim otherwise — acceptable residual, note it in the javadoc whenever the file is next touched.
- **L3: resolved, the right way** — `annotationKeys.listPrefix()` exists on the factory and the facade uses it; the drift the facade exists to prevent is now structurally closed.
- **L4: resolved** — the `\'` artifacts are gone; the IT gains the account-less `mayView` case (nice: it also asserts the guest stays OUT of the circle — both halves of the ADR-0061 boundary in one test), the circle assertion is order-insensitive, and the identity seeding is fully unique (87c18e82 + 359a5eea), which explains the earlier red run. The anonymous-review identity assertion at IT level wasn't added (unit-mocked only) — minor, unclaimed, fine.

### Remaining: §11 still carries the unqualified claim M1 corrected everywhere else

The review-level-only caveat landed in the two places that matter most — `ReviewFacade.mayView`'s javadoc (with concrete consumer guidance: forward identifiers only, let the API answer content, thread-level facade waits for a consumer per ADR-0049) and ADR-0062's decision bullet. But `docs/EXTENSION-POINTS.md` §11 at the head still reads "**exactly the annotation-listing rule**, no admin override" with no qualifier — the precise phrase the fix qualified in the ADR, unqualified in the catalogue that extension authors read first. The 359a5eea diff touched the catalogue only at §12; the commit message never claimed §11 (honest), but the resolution mapping did — mapping-vs-tree divergence, fourth of the wave, caught by the same file-at-head check as the previous three. The fix is one clause: append "review-level only — PRIVATE thread participation is not expressed (see ADR-0062)" to §11's `mayView` entry. Everything else is done; that clause closes the PR from my side.

bigpuritz (commented):
Scope: verification of commit f4b50e67 against the one remaining finding on PR #807.

**Verdict: resolved — Approve. All seven findings from both review passes are addressed; the #600/#601/#602 wave is closed from my side.**

Verified the standard way: the commit's diff contains exactly the claimed file (one line in `docs/EXTENSION-POINTS.md`), and the file fetched at the head directly confirms §11's `mayView` entry now reads "exactly the annotation-listing rule, no admin override — **review-level only**: PRIVATE thread participation is not expressed, so event-granular fan-out forwards identifiers only." The catalogue, the SPI javadoc and ADR-0062 now say the same thing in the same words — the lookup surface no longer under-qualifies what the contract qualifies.

Final state of PR #807 as reviewed: `mayView` pins one predicate through two doors (`canAccess` shared with the listing), with its review-level scope stated in all three documentation surfaces; the circle refactor is a verbatim move exercised by real-service tests, with the disabled-accounts and actor-not-excluded semantics documented; identity answers flow through the ADR-0038 resolver with benign null edges; the live-channel slot subscribes correctly for loader-time registration, tears down cleanly, isolates broken contributors, and invalidates through the factory's own `listPrefix`; the §-numbering collision with #805 carries an explicit reconciliation plan.

Wave closing note: across #805/#806/#807 the substantive findings were again of one family — a stated scope wider than the shipped code path (slot names, error-copy routing, advertisement-vs-gate, thread-level scoping, late registration) — and every one closed by narrowing the claim or widening the code, never by quietly leaving the gap. The file-at-head verification habit earned its keep four times this wave; keep it.

Issue #461, opened 2026-07-11
Audit and improve responsive behavior across device sizes


bigpuritz:
## Why

qnop-ui has grown desktop-first: the review workspace is a fixed-height split layout (viewer + resizable annotation panel), the dashboard is a two-column cockpit, and admin surfaces assume generous width. Some responsive handling exists (the review split stacks into a column on `xs`, MUI breakpoints here and there), but it has never been audited systematically — and reviewers do open documents on laptops with small screens, tablets, and occasionally phones.

## Goal

Audit every surface across common device sizes, fix what breaks, and define which experiences are fully supported vs. deliberately degraded — with visual-regression coverage so layouts don't silently regress.

### 1. Audit matrix (produce findings first)

Walk each surface at **320, 375, 768, 1024, 1440, 1920 px** (portrait and landscape where it matters):

- Login / register / password reset
- Dashboard (two-column cockpit, card scrollers, activity feed)
- Reviews list + review hub head (badges, due date, participants, actions)
- **Document review page**: viewer toolbar (version picker, page nav, tools, zoom, view-mode switch), viewer + panel split, panel resizer, re-anchoring banner, filter bar, annotation cards, comment threads, composers
- Focus mode: drawer, floating annotation card, floating composer, re-attach hint pill (#457)
- Tasks board (columns!) + task drawer, version compare
- Profile, user pages, admin surfaces (tables, forms, settings)
- App shell: sidebar, header, search, toasts

For each: no horizontal page overflow, no clipped/overlapping controls, readable type, tap targets ≥ 44px on touch sizes, usable menus/popovers.

### 2. Known suspects to verify explicitly

- **Viewer toolbar density** — many controls in one row; likely needs wrapping/priority+overflow menu below ~768px.
- **Fixed-height workspace** — the `md`-up fixed split vs. small-screen stacking: check the `xs` column layout is actually workable (panel reachable, viewer min-height sensible).
- **Resizers** (panel splitter, drawers, focus composer) on touch: drag works or a sensible fallback applies.
- **Tables and wide content** (admin, tasks list rows, compare view): scroll within their own container, never the page body.
- **Floating overlays** (focus card, composer popper, emoji picker, hover previews): stay within viewport on small screens; hover-only affordances need touch equivalents.
- **Dashboard columns** — 50/50 split should collapse gracefully; card internal scrollers on short viewports.
- **Toasts/banners** on narrow screens (multi-line, dismiss reachable).

### 3. Decide and document the support tiers

[...]

PR #808, opened 2026-08-31
feat(ui): land the approved Playwright responsive net on main (#725)

### LINKED ISSUE #725: Playwright visual-regression net for responsive layouts

### LINKED ISSUE #772: Viewer toolbar's trailing control group does not wrap at 320 and 375 px
Found by the responsive regression net (#725) on the review workspace: at **320 px** and **375 px** — split view and focus mode alike — the `main` container scrolls to **441 px**. The page body stays clean — the shell's `overflow: auto` below `md` swallows it (see finding 3 in `docs/RESPONSIVE.md`) — which is exactly why the net checks `main` too.

### Cause

`ViewerToolbar` wraps as a whole (`flexWrap: 'wrap'` on the Paper), but its trailing group — tool toggle, zoom, Panel/Focus switch, page navigation — is one `Stack direction="row"` with `ml: 'auto'` and no wrap of its own. That row measures 412 px, so once the viewport is narrower than that the group overflows instead of breaking.

Measured elements (320 px, split view): the trailing `Stack` spans x = 29…441; the Panel/Focus `ToggleButtonGroup` sits at 265…393, the page-nav button at 405…441.

### Fix

Let the trailing group wrap (or split it into two groups that wrap independently: tools + zoom, view switch + page nav), and keep `ml: 'auto'` on the last one so the right alignment survives on wide screens. Then flip the `test.fail()` marker in `qnop-ui/e2e/responsive.spec.ts` that references this issue — the net fails on an unexpected pass, so it cannot be forgotten.

Part of #461; the audit named the toolbar density as the prime suspect and this is the first measurement of it.

bigpuritz:
Closes #725 (from the #461 audit).

**Why this PR exists — third instance of the stacked-squash pattern:** PR #777 (the reviewed, green responsive net) was merged into `feat/issue-724-header-touch-targets`, and that branch's later squash to `main` predated the merge — so the net never reached `main` (no `qnop-ui/e2e/`, no Playwright CI step there today). This PR cherry-picks the four original commits onto current `main`; they applied **conflict-free**, and the tree diff against the approved #777 head shows only later main-side churn.

## What lands (as reviewed in #777)

- Playwright infra: sign-in fixture against the smoke-seeded data, run inside the official Playwright image against a preview build (`scripts/e2e-docker.sh`) so baselines are pixel-stable across CI and dev machines; clock frozen relative to the seed.
- **Overflow assertion** (`scrollWidth <= clientWidth`) on every audited surface at six widths + header touch-target checks — the image-free guard.
- **Screenshot baselines** at 320/768/1024/1440 for login, dashboard, reviews, workspace, tasks.
- CI wiring in the smoke job incl. failure report artefact; baseline update via `pnpm test:e2e:update`, reviewed through the PR image diff (documented in `docs/RESPONSIVE.md`).

## Known risk, called out

The baselines were shot on the Aug-29 UI; since then ~43 commits merged (wrap fixes #772–#775/#786, slot PRs, MUI patch bumps). The 0.5 % pixel tolerance may or may not absorb that — **if the visual step fails, the baselines need one `pnpm test:e2e:update` run (Docker required, not available in this environment) committed to this branch**; the CI artefact will show the diffs either way.

## Test plan
- [x] Cherry-picks conflict-free; tree verified against the approved head
- [x] FE `tsc`/`eslint`/`prettier` green locally
- [ ] CI green — including the new responsive step in the smoke job


bigpuritz (commented):
## Independent agent review
Scope: PR #808 as (1) a re-landing vehicle — tree identity verified against the approved `feat/issue-725-e2e-responsive` head — and (2) a first substantive review of the net itself, since #777 predates the review rule.

**Verdict: the landing mechanics are verified clean, and the net itself is well built — but one HIGH must land before merge: the spec's expected-failure markers are stale against current main and will fail deterministically, a failure mode the PR's risk note does not cover.**

### (1) Tree identity — verified

- `git diff origin/feat/issue-725-e2e-responsive <head>` touches 152 files and **zero** of them are net files: no `qnop-ui/e2e/**`, no `playwright.config.ts`, no `scripts/e2e-docker.sh`, no baseline PNG differs from the approved head. The 152 are exactly the known main-side churn: the seam wave (#790–#807 landings), the a11y test wave, the wrap fixes, Renovate, and their docs.
- The four sensitive files: `ci.yml`, `docker-compose.smoke.yml` and `vite.config.ts` are **byte-identical** to the approved branch head; `qnop-ui/package.json` differs only by main's Renovate bumps (react-query 5.102.8, axios 1.20.0, etc.), with the net's `test:e2e*` scripts and pinned `@playwright/test` riding on top. The main-vs-head delta shows the e2e block added exactly once in the smoke job (own pnpm/node/java setup — intentional, that job had none), the refresh-rate-limit env once, the proxy refactor with no main content lost. Nothing dropped, nothing duplicated. One cosmetic note: the cherry-picked `ci.yml` re-flips main's `'21'`/`'24'` quoting to `"21"`/`"24"` on ~8 untouched lines — the branch's reformat riding along; harmless.
- The `QNOP_AUTH_RATE_LIMIT_REFRESH_MAX_ATTEMPTS: '1000'` smoke-only relaxation binds correctly (`@ConfigurationProperties(prefix = "qnop.auth.rate-limit")` with a nested `Limit.maxAttempts` — relaxed binding maps the env var), and the rationale (a page-load-per-test net vs. a 30/min refresh cap) is sound and confined to the test stack.

### HIGH — the three `test.fail` markers are now wrong, and the failure is deterministic, not tolerance-dependent

`responsive.spec.ts` marks ~10 test instances as expected failures for issues #772 (workspace/focus at ≤375), #773 (dashboard/storage-consistency at 320) and #774 (review routes at 1024) — with the explicit design that "Playwright then fails on the unexpected pass, so the marker cannot outlive the bug". That design now fires: **all three issues are CLOSED and their fixes are verifiably on main** (`3e6bb152` toolbar wrap, `99e6069c`/`38e9136b` KPI rows, plus the #774 head fix — all in the 152-file churn this PR absorbs). Assuming the fixes cure the overflow at those widths, every marked test fails on unexpected pass the moment CI reaches the net — before any screenshot is compared. The PR's "known risk" note frames the exposure as "the 0.5 % pixel tolerance may or may not absorb" the UI drift; the marker failures are binary and independent of tolerance, and the note doesn't mention them. Fix in this PR: delete the three `test.fail` blocks (pure code, no Docker needed) — or, if any overflow genuinely persists at those widths, re-scope the marker to what still fails and reference a new issue. The currently-running CI will demonstrate which.

### MEDIUM — the baseline staleness is near-certain for the wrap-fixed surfaces, not "may or may not"

The same wrap fixes *moved boxes* on exactly the baselined surfaces: the viewer toolbar (workspace-320), the KPI strip (dashboard-320/768), the review head (workspace-1024). Layout changes are what `maxDiffPixelRatio: 0.005` exists to catch, so treating a baseline refresh as a contingency undersells it — plan the `pnpm test:e2e:update` commit (Docker required) as part of landing, alongside the marker removal, and review the refreshed PNGs in the PR image diff as `docs/RESPONSIVE.md` prescribes.

### (2) The net's substance — a real pass, and it holds up well

- **Fixture/auth flow:** worker-scoped signed-in context with the rotating-refresh-cookie constraint understood (no replayed storage state — the comment explains why); per-test `localStorage.clear()` init script so device-local UI state can't leak between tests; the sign-in helper names the failure (rate limit, wrong seed) instead of timing out; one API login per run in `global.setup.ts` plus one browser sign-in per worker, consciously budgeted against ADR-0027's 10/min cap.
- **Clock discipline is the standout:** freezing the clock *relative to the seeded review's creation* (+1 h) instead of a calendar date is exactly right for relative-time UIs over re-seeded data, and `page.clock.setFixedTime` runs before every `goto`.
- **The overflow assertion is honest about the shell:** it checks `main` as well as `documentElement`/`body`, closing precisely the masking hole `docs/RESPONSIVE.md` §3 warns about (the sub-`md` `overflow: auto` container swallowing body overflow) — the doc's lesson made it into the code.
- **Touch targets:** the `::before` overlay measurement matches how #724 actually grew the hit areas; labels come from `aria-label` for readable failures.
- **Stability:** `settle()` waits for fonts, all indeterminate indicators (including the hand-rolled, value-less progressbar) and a 3-sample scroll-size plateau — unbounded loop, but capped by the 60 s test timeout; the workspace additionally waits for the PDF canvas. One retry for the documented blank-tab flake is proportionate.
- **Pixel determinism:** the suite runs only inside the version-pinned official Playwright image (`e2e-docker.sh` derives the tag from the installed `@playwright/test` — no drift between dep and image), single Linux-Chromium project, `snapshotPathTemplate` without platform suffixes (matches the 20 committed PNGs), `animations: disabled`, `caret: hide`, and a `vite preview` production build rather than dev-server transforms. The 0.5 % tolerance rationale (antialiasing passes, a swapped spinner does not) is well argued.
- **CI wiring:** the net runs in the smoke job after seeding *and* after the OIDC smoke (so the login baselines deliberately include the registered provider — the d8ad0dbd fix), report artifact on failure with traces, 14-day retention.

### LOW

- **L1 — Worker count vs. the login cap:** no `workers:` cap in the config; a beefy dev machine (>10 workers) spends more browser sign-ins than ADR-0027's 10/min allows and the run dies at setup — with a named error, but still. Cap workers (e.g. 8) or note the limit in the script header.
- **L2 — Local visual runs silently require the OIDC smoke:** the login baselines include the provider button, so `scripts/e2e-docker.sh` against a stack that ran only `smoke-test.sh` fails the login screenshots. The prerequisites comment names the smoke seed but not the OIDC registration — one line there or in RESPONSIVE.md.

The re-landing is exactly what it claims; the net itself is thoughtfully engineered — the clock-relative-to-seed trick, the shell-aware overflow check and the dep-pinned image are all better than typical Playwright setups. Fix the stale markers (and pair the baseline refresh), and this is ready.

bigpuritz:
Review addressed (mapping written from the staged-file list):

- **HIGH (stale markers):** landed in the prior commit — #772/#773/#774 markers removed; replaced by `test.fail` markers for the net's first real catch, filed as #809 (reviews list at every width incl. 1920, storage consistency 375/768/1440, scheduler 320, workspace 1440 — CI measured wider than the first filing, markers match the measurement).
- **MEDIUM (baselines):** deliberate decision instead of an update run — the failing baselines are exactly the #809 surfaces (workspace, tasks), and freezing new baselines on a page that scrolls sideways would enshrine the bug. Both are expected-to-differ against #809; the one `pnpm test:e2e:update` rides with that fix. (Login/dashboard/reviews baselines held within tolerance — the staleness was narrower than feared.)
- **LOW (workers):** `workers: 4` in the config with the ADR-0027 rationale.
- **LOW (OIDC prerequisite):** named in the README beside `test:e2e`.

bigpuritz:
Marker strategy corrected after the second CI run: the #809 overflows **flicker between runs** (28 expected-fails flipped to unexpected-pass while two new surfaces went red) — `test.fail` flaps on flaky ground, so the #809 matrix (union of both runs, dashboard@320 and admin-teams@1920 included) moved to `test.fixme` in both specs. The flakiness itself is documented on #809 as a diagnostic lead. 🤖 Co-Author: Claude (Fable 5) via Claude Code

bigpuritz (commented):
Scope: verification of commits 84c40ba9 and 9c3a0bd7 against my four findings, plus analysis of the now-settled CI run (smoke job red, 9 failed / 1 flaky / 118 passed).

**Verdict: all four of my findings are genuinely resolved as mapped — but CI settled red, and the failure breakdown shows the marker scoping was calibrated against a measurement polluted by the login rate limit. One remaining HIGH with a three-part fix; the loop stays open.**

### My findings — resolved, and the mapping matches the tree

- **HIGH (stale markers):** the #772/#773/#774 markers are gone, replaced by #809 markers — and issue #809 is a model bug report (measurements, candidates, repro, the marker-removal discipline written in).
- **MEDIUM (baselines):** resolved as expected-to-differ against #809 — **and I withdraw my update-now suggestion in favor of this**: it predated knowing about #809, and refreshing workspace/tasks baselines now would enshrine a sideways-scrolling layout as the reference. The unexpected-pass discipline covers every path to the fix.
- **LOW ×2:** `workers: 4` with the ADR-0027 rationale; README names the OIDC-smoke prerequisite. Both at head.

### HIGH — the settled CI decomposes into three causes, and the marker set matches none of them

Reading the failed log, the nine failures are not one story:

1. **A genuine new catch, unmarked (2 failures).** `dashboard` at 320 overflows `main` — `scrollWidth 335 > clientWidth 320`, a real 15 px overflow — and the `dashboard-320` screenshot mismatches. Neither is in #809's scope (the old #773 marker was removed because the KPI fix landed, but *something* on the dashboard still — or again — overflows at 320 with the smoke data). This needs adding to #809 (or its own issue) plus responsive **and** visual markers.
2. **Six unexpected passes — the widened scope was mis-calibrated.** `scheduler@320`, `storage@375/768/1440`, `reviews-list@1440/1920`, `workspace@1440` all *pass* now. For reviews-list the geometry alone says the wide-width markers were wrong: a fixed ~973 px element cannot overflow a 1440/1920 viewport — it overflows exactly at ≤1024, which is where the marked tests still fail as expected. The likeliest source of the phantom wide-width "measurements": **a `test.fail`-marked test that dies on the sign-in rate limit also counts as an expected failure** — so rate-limit noise inside marked tests silently reads as "still overflowing", and the widening commit trusted that reading. Issue #809's own table (scheduler/storage rows) should be re-verified against a clean run for the same reason.
3. **The rate limit is still corrupting runs (1 flaky + 4 sign-in failures in the log).** `workers: 4` capped the *initial* sign-ins, but Playwright restarts a worker after each test failure and every restart signs in again — a failing run cascades past ADR-0027's 10/min. Until this noise is gone at the source, every marker calibration is suspect. The same trick already applied for refresh tokens is the fix: raise `QNOP_AUTH_RATE_LIMIT_LOGIN_MAX_ATTEMPTS` in `docker-compose.smoke.yml` (test stack only, like the refresh raise this PR already carries).

**Suggested order, so the next run is the calibrating one:** (1) raise the login cap in the smoke stack; (2) re-scope the responsive markers to the deterministic set — reviews-list ≤1024 stays, drop scheduler/storage/workspace wide-width markers; (3) add dashboard@320 to #809 with responsive + visual markers (and correct #809's table to the clean measurement). Then one CI run either lands green or tells the truth for the first time.

One thing worth saying plainly: behind the red X, the net is doing exactly its job — it caught the reviews-list regression, caught a dashboard regression nobody marked, and its unexpected-pass discipline is what exposed the mis-calibration. The remaining work is giving it a quiet enough environment to be believed.

bigpuritz:
Both interim MEDIUMs addressed ahead of the final: the admin-teams@1920 union cell was a sign-in rate-limit artifact and is excluded, and the smoke stack raises the login cap alongside the existing refresh raise — sign-in noise can no longer masquerade as layout failures. The probe-vs-layout diagnostic lead is recorded on #809. 🤖 Co-Author: Claude (Fable 5) via Claude Code

bigpuritz (commented):
Scope: verification of commits 84296983 and 72786024 against my follow-up findings, and the settled CI on head 72786024.

**Verdict: Approve — CI is fully green (Smoke incl. the Playwright net, 6m20s), all findings across the passes are resolved, and the #725 review loop is closed from my side.**

- **The fixme decision: sound, and I endorse it over my original framing.** Two runs with near-disjoint failure sets proved the `test.fail` ground doesn't exist for the #809 cells; `fixme` is the honest tool for flickering measurements, with the discipline parked in #809's union-as-checklist. The cost is real — ~40 dark cells until #809 is fixed — so #809 should be treated as blocking the net's value, not backlog.
- **The rate-limit artifact cell: removed** (`admin-teams@1920` out of the union, condition line only — verified in the diff), and **the noise source is closed at the stack**: `QNOP_AUTH_RATE_LIMIT_LOGIN_MAX_ATTEMPTS: '1000'` beside the refresh raise, test-stack-only, with the worker/retry-cascade rationale. This run — the first with a quiet measurement environment — settling green with zero sign-in failures is itself evidence the calibration is finally trustworthy.
- **The probe-vs-layout diagnostic lead is recorded on #809** — when that issue is picked up, reading the run-1 traces for *which element* overflowed (the assertion records per-element scrollWidth) should be step one, since a flicker that appears on geometrically impossible cells points at the measurement racing async content, not at CSS.

Closing tally for this PR: the re-landing was verified content-identical to the approved #777 head modulo main churn; the net's substance passed a real first review (clock-relative-to-seed, shell-aware overflow probe, dep-pinned image — all better than typical); and along the way the net caught one genuine regression wave (#809), had its markers mis-calibrated by rate-limit noise, and ended with both the noise source and the calibration fixed. That is a rough landing that ended in exactly the right place: a green net that can be believed, one honest issue carrying everything it skips, and the unexpected-pass discipline ready to resume the moment #809's fix removes the fixme block.

Issue #668, opened 2026-08-01
Code-scanning: 12 dependency advisories (Netty, pgjdbc, jackson-databind)


bigpuritz:
## What

Twelve open code-scanning alerts, all from the Trivy SBOM scan (no CodeQL findings):

| Package | Installed | Alerts | Fixed in |
|---|---|---|---|
| `io.netty:netty-codec-http` / `-http2` / `-compression` | 4.2.15.Final | 10 (4 high, 6 medium) | 4.2.16.Final |
| `org.postgresql:postgresql` | 42.7.11 | 1 high (CVE-2026-54291) | 42.7.12 |
| `tools.jackson.core:jackson-databind` | 3.1.4 | 1 medium (CVE-2026-59889) | 3.1.5 |

## Assessment

**None of the three is exploitable in qnop as it stands** — which is worth stating precisely, because it changes what the right fix is rather than whether to fix.

- **Netty (10 alerts).** Reached only through `software.amazon.awssdk:netty-nio-client`, the AWS SDK's *asynchronous* transport. `S3Configuration` builds the synchronous `S3Client`, which runs on `apache-client`; the Netty transport is never instantiated. The advisories are in HTTP/2 and SPDY decoders, and qnop parses no HTTP with Netty at all — the server is Tomcat, outbound calls use the JDK client and Apache.
- **pgjdbc (CVE-2026-54291, high).** A `channelBinding=require` connection can be silently downgraded to SCRAM without channel binding, losing its MITM protection. qnop's JDBC URL sets no `channelBinding`, so the guarantee is never requested and cannot be lost. It matters for an operator who *does* set it — and they would be relying on a promise the driver quietly breaks.
- **jackson-databind (CVE-2026-59889, medium).** `@JsonView` is not applied to `@JsonUnwrapped` properties on deserialisation. Neither annotation appears anywhere in this codebase.

## Fix

- **Drop the Netty transport entirely** rather than bumping it. It is unused, it brings a full HTTP/2 + SPDY codec stack, and every future Netty advisory would land in this repo's alert list for code that cannot run. An unused protocol parser is attack surface that earns nothing. (Trade-off worth naming: `S3AsyncClient` will not start without a transport — anything that wants it should add one deliberately.)
- **Pin pgjdbc to 42.7.13 and Jackson 3 to 3.1.5**, both ahead of the Spring Boot 4.1.0 BOM, until a Boot release carries them.

🤖 Co-Author: Claude (Opus 5) via Claude Code

Issue #264, opened 2026-07-01
build: align Renovate minimumReleaseAge with pnpm 11's supply-chain karenz


bigpuritz:
## Problem

pnpm 11 enforces a supply-chain **minimum release age** (~24 h) on `pnpm install --frozen-lockfile` — freshly published packages are rejected (`ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION`). It is a pnpm-11 default (not configured in the repo).

Renovate has **no** matching `minimumReleaseAge`, so it opens dependency PRs the moment an update exists — including versions younger than pnpm's karenz. Those PRs then fail the Frontend CI until the package matures (e.g. #263: `lucide-react@1.23.0`, published <24 h ago). This recurs for essentially every npm-deps PR and creates noise/red PRs.

## Fix

Set Renovate's `minimumReleaseAge` in `.github/renovate.json` to a value at/above pnpm's karenz, so Renovate only proposes versions that CI will accept. Keeps the pnpm supply-chain protection intact; just stops Renovate from racing ahead of it.

## Notes

- Could be promoted to the org preset (`qnophq/.github`) since the pnpm-11 default affects every pnpm repo; this PR fixes it repo-locally for qnop.
- Existing open deps PRs (#263) regenerate correctly on the next Renovate run once this merges, or self-heal when their packages mature.

🤖 Co-Author: Claude (Opus 4.x) via Claude Code

PR #818, opened 2026-09-06
fix(ci): re-green the three gates that turned red under every Renovate PR

### LINKED ISSUE #817: CI red on main and every Renovate PR: pnpm audit (browserslist), Trivy SBOM (Tomcat 11.0.24), OIDC smoke (curl 8.22 PSL cookies)
Every open Renovate PR (#813–#816) is red, and so is `main` itself (CI run 33558461557 on f826e735). None of the failures is caused by the dependency each PR bumps — three independent gates have gone red underneath them, two of them purely by the passage of time:

**1. Frontend gate — `pnpm audit --audit-level high`** (fails on `main` and on every PR)
`browserslist` 4.28.2 arrives transitively via `eslint-plugin-react-hooks › @babel/core › @babel/helper-compilation-targets`. Two high advisories were published against `<= 4.28.6` (GHSA-c83g-rgw3-j3cx unbounded cache growth, GHSA-73wf-gq98-2v4g prototype write via untrusted `browserslist-stats.json`); the fix is 4.28.7. Neither code path is reachable from a lint-time dependency, but the gate is intentionally unconditional (#195). PR #816 additionally hit a transient timeout against the npm advisory endpoint — a rerun, not a fix.

**2. Security workflow — Trivy SBOM scan** (fails on the three PRs run after 2026-09-03; `main`'s last Security run predates the advisories)
`tomcat-embed-core` 11.0.24, as managed by the Spring Boot 4.1.1 BOM, carries three CRITICAL advisories: CVE-2026-65182 (security-constraint bypass), CVE-2026-65905 (DIGEST replay authentication bypass), CVE-2026-68525 (FORM authentication bypass). All three are fixed in 11.0.25, which is on Maven Central but not yet in any Boot release.

**3. CI smoke — OIDC login sidecar** (fails on the three PRs run after 2026-09-03)
`scripts/smoke-oidc.sh` dies silently between "login form parsed" and "authenticated at Keycloak"; Keycloak logs `LOGIN_ERROR … error="cookie_not_found"`. The sidecar installs curl from the Alpine 3.24 repository at run time, and curl **8.22.0** landed there on 2026-09-02. Its release notes carry two cookie changes — *"cookies set for an exact PSL domain is host-only"* and *"refuse to load cookies set against a PSL domain"*. A single-label hostname such as `keycloak` is a public suffix under the PSL wildcard rule, so the `AUTH_SESSION_ID` cookie Keycloak sets on the login page is dropped when the next `curl` invocation reloads the cookie jar, and the credential POST arrives without it. Reproduced in isolation: with curl 8.22.0 a jar round-trip keeps cookies for `localhost` and `keycloak.internal` but not for `keycloak`. Nothing in the app or the Keycloak realm changed.

**Fix (one PR)**
- Bump `browserslist` in `qnop-ui/pnpm-lock.yaml` to ≥ 4.28.7 (transitive; lockfile only).
- Pin Tomcat 11.0.25 ahead of the Boot 4.1.1 BOM via a version-catalog constraint on the three `tomcat-embed-*` artifacts, following the #668 pattern (comment says why, Renovate keeps it moving, drop when Boot catches up).
- Make the OIDC smoke carry Keycloak's cookies across the login hop explicitly instead of relying on the jar for a single-label host, so the harness no longer depends on which curl Alpine ships that morning.

After the fix lands on `main`, the Renovate PRs need a rebase (their "rebase/retry" checkbox) to pick it up.

🤖 Co-Author: Claude (Fable 5.1) via Claude Code

bigpuritz:
## Summary

Every open Renovate PR (#813–#816) is red, and so is `main` (CI run 33558461557). None of it is the dependency being bumped — three gates went red underneath them, two purely by the passage of time. One PR, three independent minimal fixes:

- **Frontend — `pnpm audit --audit-level high`.** `browserslist` 4.28.2 (transitive via `eslint-plugin-react-hooks › @babel/core`) picked up two high advisories (GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g), fixed in 4.28.7. Lockfile-only move to 4.28.9 with its data packages; `package.json` untouched.
- **Security — Trivy SBOM gate.** `tomcat-embed-core` 11.0.24 (Spring Boot 4.1.1 BOM) carries three CRITICAL advisories — CVE-2026-65182, CVE-2026-65905, CVE-2026-68525: security-constraint, DIGEST-replay and FORM authentication bypasses — fixed in 11.0.25, which no Boot release carries yet. Pinned ahead of the BOM the #668 way: a `tomcat` version in the catalog, three `tomcat-embed-*` coordinates on it, and a `constraints {}` block in `qnop-app` so the starter still decides *that* Tomcat is on the classpath while the pin only lifts the version and core/el/websocket move together. Renovate keeps it moving; drop when Boot catches up.
- **CI smoke — OIDC sidecar.** The sidecar installs curl from the Alpine 3.24 repo at run time; curl **8.22.0** landed there on 2026-09-02 with two cookie changes (*cookies set for an exact PSL domain are host-only*; *refuse to load jar cookies set against a PSL domain*). A single-label hostname like `keycloak` is a public suffix under the PSL wildcard rule, so the `AUTH_SESSION_ID` cookie from the login page was silently dropped between two curl invocations and Keycloak answered the credential POST with `LOGIN_ERROR cookie_not_found`. Reproduced in isolation (jar round-trip keeps cookies for `localhost` and `keycloak.internal`, drops them for `keycloak`). The script now carries Keycloak's `Set-Cookie` values across the login hop explicitly and keeps the jar for the `localhost` hops, which curl special-cases. Also: `location()` no longer turns a missing `Location` header into a silent `set -e` exit — the caller's `case` reports what came back.

Nothing in the app, the Keycloak realm, or the compose topology changes.

## Test plan

- [x] `corepack pnpm install --frozen-lockfile` + `pnpm audit --audit-level high` → *No known vulnerabilities found*
- [x] `./gradlew :qnop-app:dependencies --configuration runtimeClasspath` → all three `tomcat-embed-*` resolve `11.0.24 -> 11.0.25`
- [x] `./gradlew clean` + `./gradlew build :qnop-app:cyclonedxBom --stacktrace`, then Trivy (`aquasec/trivy` container) on the SBOM with `--severity CRITICAL --exit-code 1`
- [x] Smoke stack locally (`docker compose -f docker-compose.smoke.yml up -d --build`, app image built from this branch): `scripts/smoke-test.sh` seeds and passes its first API checks, then stops at a macOS-only `date -d` incompatibility of the script (GNU date; CI is Ubuntu) — unrelated, left alone.
- [x] `docker compose -f docker-compose.smoke.yml run --rm smoke-oidc` against that seeded stack — the sidecar installs today's Alpine 3.24 curl **8.22.0**, i.e. exactly the curl that broke CI → *authenticated at Keycloak … ALL OIDC SMOKE CHECKS PASSED*
- [ ] CI green on this PR; then rebase the Renovate PRs (#813–#816) via their checkbox

Closes #817

🤖 Co-Author: Claude (Fable 5.1) via Claude Code

https://claude.ai/code/session_01SrEcYhdEbA1Bc3BoR76wzH

### DIFF
--- a/qnop-app/build.gradle.kts
+++ b/qnop-app/build.gradle.kts
@@ -103,6 +103,16 @@ dependencies {
     implementation(project(":qnop-api:qnop-api-endpoint")) // generated Spring interfaces (+ DTOs transitively)
 
     implementation(libs.spring.boot.starter.web)
+    // Embedded Tomcat ahead of the Boot BOM (issue #817; the `tomcat` pin in the
+    // version catalog names the advisories). Constraints, not dependencies: the
+    // starter still decides *that* Tomcat is on the classpath, this only lifts the
+    // version — Gradle resolves the higher of the BOM's and this one — and all
+    // three tomcat-embed artifacts move together. Drop when Boot catches up.
+    constraints {
+        implementation(libs.tomcat.embed.core)
+        implementation(libs.tomcat.embed.el)
+        implementation(libs.tomcat.embed.websocket)
+    }
     implementation(libs.spring.boot.starter.actuator)
     // Prometheus scrape registry for the /actuator/prometheus endpoint (issue #348). Runtime-only:
     // Boot auto-configures it; the metrics/health code uses micrometer-core (via actuator) directly.
--- a/qnop-ui/pnpm-lock.yaml
+++ b/qnop-ui/pnpm-lock.yaml
@@ -1276,8 +1276,8 @@ packages:
     resolution: {integrity: sha512-BLrgEcRTwX2o6gGxGOCNyMvGSp35YofuYzw9h1IMTRmKqttAZZVU67bdb9Pr2vUHA8+j3i2tJfjO6C6+4myGTA==}
     engines: {node: 18 || 20 || >=22}
 
-  baseline-browser-mapping@2.10.37:
-    resolution: {integrity: sha512-girxaJ7WZssDOFhzCGZTDKoTa1gk6A1TbflaYTpykLJ4UU9Fz9kx1aREM8JCuoVHbL8X8T/mJg7w2oYSq72Oig==}
+  baseline-browser-mapping@2.11.21:
+    resolution: {integrity: sha512-uh8vpY/1/YyFkunIDFH/12p7/7VdPKA1hejMVEbdkEaWnUz0Hesvx5EbiU6XxjyHZIOju+ZMbQJkRh+es3/spQ==}
     engines: {node: '>=6.0.0'}
     hasBin: true
 
@@ -1292,8 +1292,8 @@ packages:
     resolution: {integrity: sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==}
     engines: {node: 20 || >=22}
 
-  browserslist@4.28.2:
-    resolution: {integrity: sha512-48xSriZYYg+8qXna9kwqjIVzuQxi+KYWp2+5nCYnYKPTr0LvD89Jqk2Or5ogxz0NUMfIjhh2lIUX/LyX9B4oIg==}
+  browserslist@4.28.9:
+    resolution: {integrity: sha512-EWazOblFYUvlGZcfGhPUPmYh3nikUxBVb+y9MJun5f3hBi812X+8MSQTujLBtgK3cf51fJWbWfOjyeO954d+Eg==}
     engines: {node: ^6 || ^7 || ^8 || ^9 || ^10 || ^11 || ^12 || >=13.7}
     hasBin: true
 
@@ -1313,8 +1313,8 @@ packages:
     resolution: {integrity: sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==}
     engines: {node: '>=6'}
 
-  caniuse-lite@1.0.30001799:
-    resolution: {integrity: sha512-hG1bReV+OUU+MOqK4t/ZWI0tZOyz3rqS9XuhOUz1cIcbwBKjOyJEJuw9ER5JuNyqxNk8u/JUVbGibBOL1yrjFw==}
+  caniuse-lite@1.0.30001810:
+    resolution: {integrity: sha512-TITQPUkaz+aVk5GL6NhOdwk1aEaNTSDPsGFWrTuhKGtjTF70jL/Oht2W4c6rXUe5fu7Ie19VIahAXHIIiWWNeg==}
 
   ccount@2.0.1:
     resolution: {integrity: sha512-eyrF0jiFpY+3drT6383f1qhkbGsLSifNAjA61IUjZjmLCWjItY6LB9ft9YhoDgwfmclB2zhu51Lc7+95b8NRAg==}
@@ -1521,8 +1521,8 @@ packages:
   easy-table@1.1.0:
     resolution: {integrity: sha512-oq33hWOSSnl2Hoh00tZWaIPi1ievrD9aFG82/IgjlycAnW9hHx5PkJiXpxPsgEE+H7BsbVQXFVFST8TEXS6/pA==}
 
-  electron-to-chromium@1.5.372:
-    resolution: {integrity: sha512-M3yhbAlilnwqC8D21t28UCDGHyitShTmmLRU/H+b74P6Ski16Nb9HONYEaVpMj/pwC7BEo5B95FpjODLCWbtfA==}
+  electron-to-chromium@1.5.422:
+    resolution: {integrity: sha512-UvA/32XqrLDdZSn7Jllo1AYNcWji/G0d5M0GTViE7KoGBiMunw3a34Sb2KO4ZZyrSEhqsxFoVhWWJshdyfKqJA==}
 
   emoji-mart@5.6.0:
     resolution: {integrity: sha512-eJp3QRe79pjwa+duv+n7+5YsNhRcMl812EcFVwrnRvYKoNPoQb5qxU8DG6Bgwji0akHdp6D4Ln6tYLG58MFSow==}
@@ -2457,8 +2457,8 @@ packages:
       encoding:
         optional: true
 
-  node-releases@2.0.47:
-    resolution: {integrity: sha512-Uzmd6LXpouKo8EUK68IjH4+E01w/hXyV3R3g/geCJo+rXLNfh1xucB+LOzYEOQPSiUK3h/xZf0cQGcSsmyL2Og==}
+  node-releases@2.0.54:
+    resolution: {integrity: sha512-YHs7BmmcsdAI5Ozuf8JZo6PT0mv2GIWC9vMfvUC3dp65M8hn7Ux8CPL+2oBI7juNuj9d0ndhTcznq2ODBps9cQ==}
     engines: {node: '>=18'}
 
   normalize-wheel@1.0.1:
@@ -3071,8 +3071,8 @@ packages:
     resolution: {integrity: sha512-gptHNQghINnc/vTGIk0SOFGFNXw7JVrlRUtConJRlvaw6DuX0wO5Jeko9sWrMBhh+PsYAZ7oXAiOnf/UKogyiw==}
     engines: {node: '>= 10.0.0'}
 
-  update-browserslist-db@1.2.3:
-    resolution: {integrity: sha512-Js0m9cx+qOgDxo0eMiFGEueWztz+d4+M3rGlmKPT+T4IS/jP4ylw3Nwpu6cpTTP8R1MAC1kF4VbdLt3ARf209w==}
+  update-browserslist-db@1.3.2:
+    resolution: {integrity: sha512-UQ+MSxlhRm1bzjhU+DcuXfjFO1FzNtqhK5+9Yvlp90ItDLk5vT932A0rFu619nf7RVS+Y/VeaUW1jaRDqZ8VJw==}
     hasBin: true
     peerDependencies:
       browserslist: '>= 4.21.0'
@@ -3370,7 +3370,7 @@ snapshots:
     dependencies:
       '@babel/compat-data': 7.29.7
       '@babel/helper-validator-option': 7.29.7
-      browserslist: 4.28.2
+      browserslist: 4.28.9
       lru-cache: 5.1.1
       semver: 6.3.1
 
@@ -4490,7 +4490,7 @@ snapshots:
 
   balanced-match@4.0.4: {}
 
-  baseline-browser-mapping@2.10.37: {}
+  baseline-browser-mapping@2.11.21: {}
 
   basic-ftp@5.3.1: {}
 
@@ -4502,13 +4502,13 @@ snapshots:
     dependencies:
       balanced-match: 4.0.4
 
-  browserslist@4.28.2:
+  browserslist@4.28.9:
     dependencies:
-      baseline-browser-mapping: 2.10.37
-      caniuse-lite: 1.0.30001799
-      electron-to-chromium: 1.5.372
-      node-releases: 2.0.47
-      update-browserslist-db: 1.2.3(browserslist@4.28.2)
+      baseline-browser-mapping: 2.11.21
+      caniuse-lite: 1.0.30001810
+      electron-to-chromium: 1.5.422
+      node-releases: 2.0.54
+      update-browserslist-db: 1.3.2(browserslist@4.28.9)
 
   call-bind-apply-helpers@1.0.2:
     dependencies:
@@ -4529,7 +4529,7 @@ snapshots:
 
   callsites@3.1.0: {}
 
-  caniuse-lite@1.0.30001799: {}
+  caniuse-lite@1.0.30001810: {}
 
   ccount@2.0.1: {}
 
@@ -4735,7 +4735,7 @@ snapshots:
     optionalDependencies:
       wcwidth: 1.0.1
 
-  electron-to-chromium@1.5.372: {}
+  electron-to-chromium@1.5.422: {}
 
   emoji-mart@5.6.0: {}
 
@@ -5958,7 +5958,7 @@ snapshots:
     dependencies:
       whatwg-url: 5.0.0
 
-  node-releases@2.0.47: {}
+  node-releases@2.0.54: {}
 
   normalize-wheel@1.0.1: {}
 
@@ -6697,9 +6697,9 @@ snapshots:
 
   universalify@2.0.1: {}
 
-  update-browserslist-db@1.2.3(browserslist@4.28.2):
+  update-browserslist-db@1.3.2(browserslist@4.28.9):
     dependencies:
-      browserslist: 4.28.2
+      browserslist: 4.28.9
       escalade: 3.2.0
       picocolors: 1.1.1
 
--- a/scripts/smoke-oidc.sh
+++ b/scripts/smoke-oidc.sh
@@ -38,8 +38,18 @@ fail() {
   echo "[oidc-smoke] FAIL: $*" >&2
   exit 1
 }
-# Location header value of a saved response (CR-stripped).
-location() { grep -i '^location:' "$1" | tr -d '\r' | sed -E 's/^[Ll]ocation:[[:space:]]*//'; }
+# Location header value of a saved response (CR-stripped). Empty when the
+# response carried none — deliberately not a pipeline failure, so the caller's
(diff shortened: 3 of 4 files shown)


bigpuritz (commented):
## Independent review of #818

I reviewed this PR without prior context, read-only (no checkout, no edits). Verdict up front: **no blocking findings**. All three fixes are correct, minimal, and consistent with the repository's conventions; every claim in the PR body I could test held up, with one wording nuance in a script comment (LOW) noted below. Per-line comments carry the details.

### What I verified, and how

**Tomcat pin (gradle/libs.versions.toml, qnop-app/build.gradle.kts)**
- `./gradlew -q :qnop-app:dependencies` for `runtimeClasspath`, `testRuntimeClasspath` and `productionRuntimeClasspath`: every path to `tomcat-embed-core`, `-el` and `-websocket` resolves `11.0.24 -> 11.0.25`. The constraint reaches the test classpath too because `testImplementation` extends `implementation`.
- Maven Central: `tomcat-embed-core` latest/release is `11.0.25`; `spring-boot-dependencies:4.1.1` manages `<tomcat.version>11.0.24</tomcat.version>`; there is no Boot 4.1.2 (404) and 4.0.8 also manages 11.0.24, so "no Boot release carries it yet" is accurate.
- GitHub Advisory DB: CVE-2026-65182 (GHSA-gcx9-497g-6cp6), CVE-2026-65905 (GHSA-9xv2-5v5q-p794), CVE-2026-68525 (GHSA-h3x4-894j-xpx5) are all `critical`, published 2026-08-26, affected package `org.apache.tomcat.embed:tomcat-embed-core >= 11.0.0-M1, < 11.0.25`, `first_patched_version 11.0.25`. `tomcat-embed-el` and `-websocket` are not listed; moving them together is still the right call (Tomcat publishes the three as one release).
- Constraint semantics: the lift works because the BOM is imported via `platform()` (non-strict constraints, highest wins). With `enforcedPlatform()` the same block would fail resolution. The repo uses `platform()` everywhere, so this is fine; worth knowing if that ever changes.
- Renovate: `[libraries]` entries with `version.ref` are exactly what the gradle manager tracks; the three coordinates share the one `tomcat` ref, so they move as a single edit. Removing the pin later is deleting the catalog block plus the `constraints {}` block. There is no packageRule group for `org.apache.tomcat.embed`, and none is needed with a shared ref.
- Pattern vs. #668 (a60e18e4): #668 put `version.ref` on directly-declared dependencies (`postgresql`, `jackson3-databind`); nobody declares Tomcat directly, so `constraints {}` is the correct mechanism here rather than a deviation. The Trivy SBOM job on this PR is green.

**browserslist lockfile bump (qnop-ui/pnpm-lock.yaml)**
- `npm view <pkg>@<ver> dist.integrity` for all six moved packages matches the lockfile hashes byte for byte.
- GHSA-c83g-rgw3-j3cx and GHSA-73wf-gq98-2v4g: both `high`, published 2026-09-01, range `<= 4.28.6`, patched `4.28.7`.
- `pnpm audit --audit-level high` on this branch's lockfile: *No known vulnerabilities found*. `package.json` untouched. Frontend CI job is green.

**OIDC smoke (scripts/smoke-oidc.sh)**
- `bash -n` passes. (shellcheck is not installed here.)
- Reproduced the root cause in a throwaway `alpine:3.24` container with `--add-host keycloak:127.0.0.1` and a busybox `nc` fake server, using the curl the Alpine repo ships today (`curl-8.22.0-r0`, libpsl 0.21.5): a `-c jar`/`-b jar` round-trip sends **no** `Cookie:` on the second hop for host `keycloak`, but does for `keycloak.internal` and `localhost`. The PR's isolation claim is exactly right, and the `localhost` special case that the remaining jar hops (steps 4 and 7) rely on is confirmed.
- curl 8.22.0 changelog carries exactly the two entries quoted ("cookies set for an exact PSL domain is host-only", "refuse to load cookies set against a PSL domain"). I did not independently verify the 2026-09-02 landing date in Alpine; I only confirmed 8.22.0-r0 is what the repo serves now.
- `cookies_of()` executed under bash with **busybox** `grep`/`tr`/`sed`/`paste` (`/usr/bin/paste -> /bin/busybox`, BusyBox 1.37.0) on a two-`Set-Cookie` response with attributes, CRLF line ends and an `=` inside a cookie value yields `AUTH_SESSION_ID=abc.def; KC_RESTART=eyJ.x=y.z`, and curl sends it verbatim as the `Cookie:` header when passed via `-b`. `paste -sd ';' -` is supported by busybox. A `;` cannot appear in a cookie-value (RFC 6265), so `s/;.*$//` is safe.
- Both helpers return empty with rc 0 on a header file without the header under `set -euo pipefail`; the old `location()` form under the same flags exits the shell with status 1 and no output, so the "silent `set -e` exit" description is accurate.
- The Smoke CI job on this PR (which runs this script in the real sidecar) is green.

**Conventions**: issue-first (#817), Conventional Commit `fix(ci):`, Conventional Branch `fix/issue-817-ci-gates`, DCO sign-off, Claude attribution in commit and PR body, SPDX headers present on all touched files, everything in English. No ADR needed; nothing here is an architecture decision.

### Findings (none blocking)

| Severity | File | Finding |
|---|---|---|
| LOW | `scripts/smoke-oidc.sh:113` | Comment says curl 8.22.0 "refuses to save or load" jar cookies for a PSL host. Verified: it **does save** them (`#HttpOnly_keycloak … AUTH_SESSION_ID` lands in the jar) and refuses to **load** them, matching the changelog entry. The fix is unaffected; the comment overstates the save half. |
| LOW | `qnop-app/build.gradle.kts:111` | Informational: the pin covers everything qnop-app resolves (shipped jar, SBOM, its tests). Module-local classpaths of `qnop-core` (only `tomcat-embed-el` 11.0.24) and `qnop-api-endpoint` (`tomcat-embed-core` 11.0.24, no tests) stay on the BOM version. Nothing ships from there, so no action; noting it so nobody is surprised by a module-level `dependencies` report. |
| LOW | `qnop-ui/pnpm-lock.yaml:1295` | Informational: `browserslist@4.28.9` was published 2026-09-04T13:39Z, about 43 h before this PR. That clears pnpm 11's ~24 h floor (CI proves it) but sits inside the repo's own 3-day Renovate quarantine from #264/ADR-0017. `4.28.8` (2026-08-08) would have satisfied the advisory with the full margin. No change needed. |
| NIT | `scripts/smoke-oidc.sh:51` | The header-name strip `s/^[Ss]et-[Cc]ookie:…//` is only partly case-insensitive while the `grep -i` before it is fully so; `s/^[^:]*:[[:space:]]*//` would make the two agree. Same pre-existing pattern in `location()`. |
| NIT | `scripts/smoke-oidc.sh:120` | A deletion cookie (`NAME=; Max-Age=0`) would be forwarded as `NAME=`. Keycloak's fresh login page sets none (Smoke is green), so this is future-proofing only. |

**Security**: the pin removes three CRITICAL container-auth bypasses from the shipped classpath. qnop authenticates through Spring Security rather than Tomcat's `security-constraint`/DIGEST/FORM authenticators, so none of the three was reachable here in practice, but the gate and the hygiene are right. The smoke change moves test-realm session cookies from a temp file into a curl argument inside a throwaway sidecar; no real credentials are involved.

**Verdict**: approve-worthy as is. The LOW/NIT items are optional polish.

🤖 Independent review by Claude (Fable 5.1) via Claude Code

scripts/smoke-oidc.sh
```diff
@@ -98,7 +108,17 @@ esac
 log "authorization request initiated"
 
 # 5) Fetch the Keycloak login page and extract the form POST action.
-curl -s -c "$JAR" -b "$JAR" -o "${TMP}/login.html" "$auth_url"
+#    Keycloak binds the pending login to cookies it sets on this response
+#    (AUTH_SESSION_ID, KC_RESTART) and expects them back on the credential POST.
+#    They travel explicitly, not via the cookie jar: curl 8.22.0 refuses to save
```
bigpuritz: **LOW (comment accuracy)** — "refuses to save or load" overstates the save half.

Verified in an `alpine:3.24` container with the repo's current `curl-8.22.0-r0` (libpsl 0.21.5) and `--add-host keycloak:127.0.0.1`: after `curl -c jar http://keycloak:…` the jar **does** contain the cookie:

```
#HttpOnly_keycloak	FALSE	/realms/qnop	FALSE	0	AUTH_SESSION_ID	abc
```

What fails is the next invocation's `-b jar`: no `Cookie:` header is sent for host `keycloak` (it is for `keycloak.internal` and `localhost`). That matches the 8.22.0 changelog entry "refuse to load cookies set against a PSL domain" and your PR body/commit message, which both say "load". Suggest the comment say the same, e.g. "curl 8.22.0 refuses to *load* jar cookies whose host is a public suffix". The fix itself is unaffected.

scripts/smoke-oidc.sh
```diff
@@ -38,8 +38,18 @@ fail() {
...
+# response carried none — deliberately not a pipeline failure, so the caller's
+# `case` reports *what* came back instead of `set -e` killing the run silently.
+location() {
+  { grep -i '^location:' "$1" || true; } | tr -d '\r' | sed -E 's/^[Ll]ocation:[[:space:]]*//'
+}
+# Cookie header value assembled from every Set-Cookie of a saved response
+# ("name=value; name=value"), attributes stripped.
+cookies_of() {
+  { grep -i '^set-cookie:' "$1" || true; } | tr -d '\r' |
+    sed -E 's/^[Ss]et-[Cc]ookie:[[:space:]]*//; s/;.*$//' | paste -sd ';' - | sed 's/;/; /g'
```
bigpuritz: **NIT** — the `grep -i` above is fully case-insensitive, but this strip only covers `[Ss]et-[Cc]ookie`. An all-caps `SET-COOKIE:` line would survive into the `-b` string as `SET-COOKIE: name=value`. Unrealistic for Keycloak (and `location()` has the same pre-existing asymmetry), but `s/^[^:]*:[[:space:]]*//` would make the two agree without depending on the header's spelling.

For the record, I ran this exact pipeline under bash with **busybox** `grep`/`tr`/`sed`/`paste` on Alpine 3.24 (BusyBox 1.37.0; `paste -sd ';' -` is supported) against a two-`Set-Cookie` response with CRLF, attributes and an `=` inside a value: output `AUTH_SESSION_ID=abc.def; KC_RESTART=eyJ.x=y.z`, and curl sent it verbatim as the `Cookie:` header. Empty input yields an empty string with rc 0 under `set -euo pipefail`. So the portability concern is settled.

scripts/smoke-oidc.sh
```diff
@@ -98,7 +108,17 @@ esac
...
+#    Keycloak binds the pending login to cookies it sets on this response
+#    (AUTH_SESSION_ID, KC_RESTART) and expects them back on the credential POST.
+#    They travel explicitly, not via the cookie jar: curl 8.22.0 refuses to save
+#    or load jar cookies whose host is a public suffix, and under the PSL
+#    wildcard rule a single-label hostname such as `keycloak` is one — the
+#    round-trip silently dropped them and Keycloak answered the POST with
+#    LOGIN_ERROR cookie_not_found (issue #817). The jar stays in use for the
+#    app hops on localhost, which curl special-cases.
+curl -s -o "${TMP}/login.html" -D "${TMP}/b" "$auth_url"
+kc_cookies="$(cookies_of "${TMP}/b")"
```
bigpuritz, line 122: **NIT** — `cookies_of` forwards every `Set-Cookie`, including a deletion (`NAME=; Max-Age=0` becomes `NAME=`). Keycloak's login page for a fresh session sets only `AUTH_SESSION_ID` and `KC_RESTART` (and the Smoke job on this PR is green), so nothing to fix now; if you want to future-proof, dropping entries with an empty value (`grep -v '=$'` before the `paste`) would do it. Optional.

qnop-app/build.gradle.kts
```diff
@@ -103,6 +103,16 @@ dependencies {
     implementation(project(":qnop-api:qnop-api-endpoint")) // generated Spring interfaces (+ DTOs transitively)
 
     implementation(libs.spring.boot.starter.web)
+    // Embedded Tomcat ahead of the Boot BOM (issue #817; the `tomcat` pin in the
+    // version catalog names the advisories). Constraints, not dependencies: the
+    // starter still decides *that* Tomcat is on the classpath, this only lifts the
+    // version — Gradle resolves the higher of the BOM's and this one — and all
+    // three tomcat-embed artifacts move together. Drop when Boot catches up.
+    constraints {
```
bigpuritz, line 111: **LOW (informational, no action)** — scope of the constraint, verified with `./gradlew -q :qnop-app:dependencies --configuration <c>`:

- `runtimeClasspath`, `testRuntimeClasspath`, `productionRuntimeClasspath`: every path to `tomcat-embed-core`/`-el`/`-websocket` shows `11.0.24 -> 11.0.25`. So the boot jar, the SBOM the Trivy gate scans, and this module's tests all get the fix.
- `qnop-core` (`tomcat-embed-el` only) and `qnop-api-endpoint` (`tomcat-embed-core` via its own `starter-web`, no test sources) keep resolving 11.0.24 on their module-local classpaths. Nothing ships from there and none of the three CVEs touch `embed-el`, so keeping the pin local to the runnable module is the minimal and correct choice; noting it only so a module-level `dependencies` report does not surprise anyone.

Two things worth knowing about the mechanism: it works because the BOM comes in via `platform()` (plain constraints, highest wins). With `enforcedPlatform()` this block would fail resolution instead of lifting. And Renovate's gradle manager tracks `version.ref`, so the three coordinates move together on a single `tomcat` edit; removal is deleting this block plus the catalog entries. All consistent with the comment.

qnop-ui/pnpm-lock.yaml
```diff
@@ -1292,8 +1292,8 @@ packages:
     resolution: {integrity: sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==}
     engines: {node: 20 || >=22}
 
-  browserslist@4.28.2:
-    resolution: {integrity: sha512-48xSriZYYg+8qXna9kwqjIVzuQxi+KYWp2+5nCYnYKPTr0LvD89Jqk2Or5ogxz0NUMfIjhh2lIUX/LyX9B4oIg==}
+  browserslist@4.28.9:
```
bigpuritz, line 1295: **LOW (informational, no action)** — `browserslist@4.28.9` was published 2026-09-04T13:39Z, roughly 43 h before this PR. That clears pnpm 11's ~24 h supply-chain floor (the Frontend job is green) but is inside the repo's own 3-day `minimumReleaseAge` that Renovate applies as margin (#264 / ADR-0017). `4.28.8` (2026-08-08) would have satisfied both advisories (`patched 4.28.7`) with the full margin. Not asking for a change; just flagging that a hand-made lockfile bump bypasses the quarantine Renovate would have honoured.

Verified for this entry and the five data packages: `npm view <pkg>@<ver> dist.integrity` matches every lockfile hash, and `pnpm audit --audit-level high` on this lockfile reports *No known vulnerabilities found*.

scripts/smoke-oidc.sh
bigpuritz: Agreed and fixed in 663479b9: the comment now says curl 8.22.0 still writes the cookie to the jar and refuses to *load* it, which is also what the commit message and PR body say.

scripts/smoke-oidc.sh
bigpuritz: Taken in 663479b9: both helpers strip the header name with `^[^:]*:` now, so the strip is as case-insensitive as the `grep -i` in front of it — applied to `location()` too, since this PR rewrote it anyway.

scripts/smoke-oidc.sh
bigpuritz: Taken in 663479b9: `cookies_of` now drops entries with an empty value (`name=`), so a deletion cookie is no longer echoed back. Exercised under Alpine 3.24's busybox grep/sed/paste with a mixed-case `SET-COOKIE: OLD=; Max-Age=0`, an `=` inside a value and empty results under `set -euo pipefail`; the full OIDC smoke was re-run locally against the sidecar's curl 8.22.0 and passes.

qnop-app/build.gradle.kts
bigpuritz: Thanks for verifying the scope across the three classpaths — that matches the intent: only qnop-app ships, and the constraint deliberately stays module-local rather than moving into the convention plugin. No change.

qnop-ui/pnpm-lock.yaml
bigpuritz: Noted. 4.28.9 clears pnpm 11's floor (CI's `pnpm install --frozen-lockfile` passed on this PR), and the 3-day quarantine is Renovate's own scheduling margin rather than a rule for a manual advisory fix; by the time this merges it is past that window too. Leaving 4.28.9.

